PR #5, opened 2018-12-07
 protosanitizer: handle secrets in oneof and maps, gRPC-style output

pohly:
This adds the support for filtering secrets also out of oneof fields and maps, something that was left out of the initial version in order to get it reviewed and merged faster.

gRPC-style output IMHO is easier to read, but that is of course subjective.

Both enhancements are presented in the same PR because they both touch the testcases and would conflict otherwise.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -27,6 +27,14 @@
   version = "v1.2.0"
 
 [[projects]]
+  digest = "1:40e195917a951a8bf867cd05de2a46aaf1806c50cf92eebf4c16f78cd196f747"
+  name = "github.com/pkg/errors"
+  packages = ["."]
+  pruneopts = "UT"
+  revision = "645ef00459ed84a119197bfb8d8205042c6df63d"
+  version = "v0.8.0"
+
+[[projects]]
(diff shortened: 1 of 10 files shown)


pohly:
@saad-ali I leave it up to you whether
a) this should go in before tagging a v0.1.1 that then also gets vendored into the 0.4.x apps
b) this should go in after v0.1.1 and then become v0.2.0
c) not go in at all because some other solution is better
d) only one of the two commits should get merged (then I'll adapt the PR accordingly)

jsafrane:
Is this PR still necessary? I can see some sanitizer already merged.

pohly:
This PR is an enhancement of what's currently in master. It makes the code more future-proof by also filtering out secret fields in oneof fields and maps. It also makes the output a bit easier to read.

pohly:
/remove-lifecycle stale

@msau42 should we finally merge this?

pohly:
/remove-lifecycle stale

I don't know whether anyone still cares, but this is still a useful enhancement :grin:

pohly:
https://blog.golang.org/a-new-go-api-for-protocol-buffers should make this recursive removal of sensitive fields a lot easier.

PR #7, opened 2018-12-21
remove non-go code from vendor

Madhu-1:


Madhu-1:
@msau42 @saad-ali PTAL

PR #8, opened 2018-12-24
Add connection package for rpc connection logic

### LINKED ISSUE #6: Create common library for rpc connection logic
Currently all the sidecars copy paste the same connection logic

/help

darkowlzz:
This change adds `connection` package, which could be imported and
used by the sidecars.
The `CSIConnection` interface combines methods from external-attacher
and driver-registrar.

Also adds `.travis.yml` file for setting up the CI.

Fixes #6


darkowlzz:
I've removed the commit that added .travis.yaml, rebased and generated new lock file.

Gopkg.toml
```diff
@@ -5,3 +5,23 @@
...
+  name = "google.golang.org/grpc"
+  version = "1.9.2"
+
+[[constraint]]
+  name = "github.com/golang/protobuf"
+  version = "1.2.0"
+
+[[constraint]]
+  name = "github.com/golang/mock"
+  version = "v1.1.1"
```
pohly: Are there reasons why you are locking these projects to specific versions?

Usually "the latest stable release" (the default in dep) should be fine, so I would remove all of the above. If there is a specific reason to be more specific, add a constraint with a comment explaining the reason.

pohly: Also please decide where you want to update the dependencies. I'm fine with both doing it all in one commit as well as "first code change, then vendor change", but right now you have two commits which both touch `vendor` and `Gopkg.toml`.

msau42: We discussed this during the csi kubernetes meeting.  There's some concern that if these get removed, then we could see many CI failures if CI is always trying to pull the latest.  In addition, if we could see more breakages if we're always updating to the latest.

cc @saad-ali @xing-yang

xing-yang: This is not really locked to the specific version. When you specify a version without an operator, dep automatically uses the ^ operator by default.  So `version = “1.2.3”` becomes the range of >=1.2.3 and <2.0.0.   If you really want to pin the dependency to "1.2.3", you have to set `version = "=1.2.3"`.

https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md#version

pohly: Testing in the CI should always use exactly the source that is in `vendor` - that's what it is for.

I know that `dep` is interpreting this particular revision as a certain range. But my point still stands: why limit to that range? If there are reasons, then state them in the file so that others can know about them and do not accidentally change something.

msau42: @darkowlzz can you clarify if these versions were filled in automatically by the dep tool?

pohly: I discussed this on #dep-users (https://gophers.slack.com/team/UB2C0343D, not sure if that link helps). The summary is:
- it makes sense to lock onto a major version range because a major version change really should be done manually (as in, a human knows about the change and enables it by editing the `Gopk.toml` file); this is a safety precaution and the assumption is that smaller changes are okay to be done semi-automatically
- the version number should reflect what the code is known to work with, it doesn't necessarily need to be the smallest suitable version (although that would be better)

`dep init` does add constraints, but it's a bit ambiguous what version it uses. This needs to be reviewed to ensure that it matches what I said above.

With that in mind, what I'd like to see in this PR for each constrain is either a) a version number that matches what gets vendored and tested or b) a comment explaining why some lower version is okay.

darkowlzz: Hi, it has been a long time since I created this PR, so I don't exactly remember if I used `dep ensure` to add the constraints or I added them manually. But that need not be a big concern because dep will always select the latest release under the constrained major version. Defining constraints for a library is not necessary but it's good to have and it helps the users of the library with some extra information. At least that's what we (me and other dep authors), had in mind while writing dep. And the defined constraints in Gopkg.toml help in managing the dependencies of the project in a controlled manner.
I don't think it's necessary to explain why a lower version is okay when it's a minor or a patch version, unless there's some exception. For a major version number, it would be useful, but in this case, I don't think it's needed.

The way `dep init` adds constraints is by importing constraints from any existing dependency management tools. If no dependency manifest was found, it checks the GOPATH for the version of the projects that are used in the current project and uses the version of the project from GOPATH. If the project isn't available in GOPATH, it checks the project repository online and selects the latest version of that project. This is the case only for `init`. `ensure` will always fetch the latest version of a project from online repo, if no constraint is specified to `-add` flag parameter.

Upon further investigation, I found that setting `google.golang.org/grpc` to `1.9.2` seems to have come from the external-attacher project, because most of the code in this PR is taken from there. Likewise, it's probable that I took the other constraints from there as well.

I've squashed all the commits and updated all the constraints to the latest release versions, same as in the lock file. It should be good now :slightly_smiling_face:

darkowlzz: np, it's alright :hugs: . Resolving this thread.

jsafrane:
I don't think it's necessary to pull additional ~90 new files for a simple Connect() function. This PR contains lot of unrelated changes.

jsafrane:
I filled https://github.com/kubernetes-csi/csi-lib-utils/pull/11, which is much smaller.

jsafrane:
Obsoleted by https://github.com/kubernetes-csi/csi-lib-utils/pull/11
/close

PR #9, opened 2019-01-11
Correct markdown formatting.

humblec:
Signed-off-by: hchiramm <hchiramm@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,8 @@ Learn how to engage with the Kubernetes community on the [community page](http:/
 
 You can reach the maintainers of this project at:
 
-- [Slack](http://slack.k8s.io/)
-- [Mailing List](https://groups.google.com/forum/#!forum/kubernetes-dev)
+* [Slack](http://slack.k8s.io/)
+* [Mailing List](https://groups.google.com/forum/#!forum/kubernetes-dev)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,8 +8,8 @@ Learn how to engage with the Kubernetes community on the [community page](http:/
 
 You can reach the maintainers of this project at:
 
-- [Slack](http://slack.k8s.io/)
-- [Mailing List](https://groups.google.com/forum/#!forum/kubernetes-dev)
+* [Slack](http://slack.k8s.io/)
+* [Mailing List](https://groups.google.com/forum/#!forum/kubernetes-dev)
```
pohly: Why are bullets better than hyphens? Both are listed as valid markup for lists in multiple places:
https://en.support.wordpress.com/markdown-quick-reference/
https://guides.github.com/features/mastering-markdown/

This particular text fragment comes from the Kubernetes template for new repos: https://github.com/kubernetes/kubernetes-template-project

I don't think we should deviate from that.

msau42: @humblec did you use some linter tool to find these changes? Can we work on adding common fmt/linting tools to the csi-lib-utils repo so the rules are consistent across all repos?

PR #11, opened 2019-01-18
Add Connect function

jsafrane:
Add a common function to connect to a CSI driver.

Expectation is that it will be used by every CSI sidecar container so they have the same handling ~~+ the same default timeout (1 minute)~~. Sidecar container should wait indefinitely until CSI driver exposes a socket.

~~Is 1 minute the right timeout?~~

@msau42 @saad-ali @pohly @sbezverk @lpabon @vladimirvivien @davidz627 @gnufied

There is no (unit) test because there is not much to test...


jsafrane:
"Build Still Running" for 2 hours but it's marked as failed.

/retest

jsafrane:
Reworked to wait forever as agreed on Friday's CSI meeting. As result, logging was added to print something every 10 seconds.

connection/connection.go
```diff
@@ -0,0 +1,69 @@
...
+	}()
+
+	// Log error every connectionLoggingInterval
+	ticker := time.NewTicker(connectionLoggingInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			glog.Warningf("Still connecting to %s", address)
```
pohly: `glog` has been replaced in Kubernetes. We should do the same in kubernetes-csi. Can you replace with `k8s.io/klog` here and in the rest of the new code?

connection/connection.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/kubernetes-csi/csi-lib-utils/protosanitizer"
+	"google.golang.org/grpc"
+)
+
+const (
+	// Interval of logging connection errors
+	connectionLoggingInterval = 10 * time.Second
+)
+
+// Connect opens insecure gRPC connection to a CSI driver. Address must have either '<protocol>://' prefix, or be
```
pohly: Better document what protocols are valid or link to documentation about that. I know of that people have been bitten by attempts to use "tcp4" as prefix, which is not what gRPC uses (it wants `dns`).

connection/connection.go
```diff
@@ -0,0 +1,69 @@
...
+// a path to a socket file. The function tries to connect indefinitely every second until it connects.
+// The function automatically adds interceptor for gRPC message logging.
+func Connect(address string, dialOptions ...grpc.DialOption) (*grpc.ClientConn, error) {
+	dialOptions = append(dialOptions,
+		grpc.WithInsecure(),                   // Don't use TLS, it's usually local Unix domain socket in a container.
+		grpc.WithBackoffMaxDelay(time.Second), // Retry every second after failure.
+		grpc.WithBlock(),                      // Block until connection succeeds.
+		grpc.WithUnaryInterceptor(LogGRPC),    // Log all messages.
+	)
+	if strings.HasPrefix(address, "/") {
```
pohly: This doesn't handle an explicit `unix:///` prefix.

connection/connection.go
```diff
@@ -0,0 +1,69 @@
...
+
+const (
+	// Interval of logging connection errors
+	connectionLoggingInterval = 10 * time.Second
+)
+
+// Connect opens insecure gRPC connection to a CSI driver. Address must have either '<protocol>://' prefix, or be
+// a path to a socket file. The function tries to connect indefinitely every second until it connects.
+// The function automatically adds interceptor for gRPC message logging.
+func Connect(address string, dialOptions ...grpc.DialOption) (*grpc.ClientConn, error) {
```
pohly: Is it worthwhile to pass in a context? Then a caller can decide to abort after a certain time (`context.WithTimeout`) or run `Connect` in a goroutine and cancel the operation (`context.WithCancel`).

connection/connection.go
jsafrane: `Dial()` handles `unix:///` just fine without any special dialer.

connection/connection.go
jsafrane: Added link to https://github.com/grpc/grpc/blob/master/doc/naming.md

connection/connection.go
jsafrane: IMO, the point of this function is to have the same behavior in all sidecar containers. Adding optional parameters works against this unification.

connection/connection.go
jsafrane: We should move all repos to klog in one (short) burst. Is it the right time?

connection/connection.go
jsafrane: We should move all repos to klog in one (short) burst. Is it the right time?

connection/connection.go
pohly: Are you sure? https://github.com/grpc/grpc-go/issues/1741 was asking for this and then the feature request moved to https://github.com/grpc/grpc-go/issues/1911, which is still open.

I was under the impression that the custom dialer was used specifically to work around that shortcoming in gRPC. Perhaps I misunderstood. :man_shrugging: 

Looks like a good test case, don't you think? :wink:

connection/connection.go
pohly: We can have the same behavior in all sidecars by calling the function the same way, while still writing the underlying function in a more flexible way. But I get your point.

Let me add another reason for passing in a context: when discussing glog vs. klog, it was mentioned that it would be good to pass in a logger via the context instead of using the same global instance. If we had that now, a unit test could select a mock logger to test that the warning messages are printed.

But we don't have that now, so I'm fine with leaving out the context and adding it when really needed. We can deal with the API change in the sidecars on a case-by-case basis, should we do it later.

connection/connection.go
pohly: It does indeed work. Then I don't understand what those open gRPC issues are about. It looks like the custom dialer is needed because an absolute path without the `unix:///` prefix will not be treated as a Unix domain socket.

I'm still a bit wary about not always using that dialer, though. It creates two different code paths in our `Connect` function which (theoretically) could have different behavior. In practice I've not found any relevant difference, but the slow tests in PR https://github.com/jsafrane/csi-lib-utils/pull/1 now only cover the case with an absolute path. I only tried the `unix:///` variant of those by editing the test source code.

connection/connection.go
```diff
@@ -0,0 +1,71 @@
+package connection
```
pohly: I think we need the usual copyright and license boiler plate code at the beginning of this file.

connection/connection.go
jsafrane: I agree, but we should run it through CSI WG meeting. And add common logging initialization function to this library (as separate PR), so all the sidecar containers have the same cmdline arguments for logging.

connection/connection.go
jsafrane: I added code that removes custom dialer and uses unix:// for all sockets. This changes behavior of existing sidecar containers that used custom dialer in old releases, IMO it should be fine.

pohly:
In the January 25th CSI WG meeting we discussed klog and how to handle disconnects.

For klog, we agreed to make the switch, so we should start also here in this PR.

For disconnects I proposed the following:
- add an optional callback to `Connect` that gets called by the dialer before reconnecting.
- apps that don't do caching can ignore that and transparently get reconnected to the CSI driver
- apps that do caching can inside their callback:
  - exit to trigger a restart
  - invalidate their cached information and continue running

pohly:
@jsafrane let's chat on Monday to determine who does what. I'm fine with leaving it all to you or take on some of the work.

PR #16, opened 2019-02-11
Add GetDriverName function

jsafrane:


msau42:
Do we also want to move some common methods like get controller capabilities.  Probe? Or is that too simple to be worth putting into a library?

jsafrane:
> Do we also want to move some common methods like get controller capabilities. Probe?

See https://github.com/kubernetes-csi/csi-lib-utils/pull/17

PR #17, opened 2019-02-14
Add Get*Capabilities and Probe common functions

jsafrane:
These functions can be used by most of our sidecar containers and it makes sense to move them to util library.

Open question: should `Probe()` wait until the driver is ready? How long? Various sidecars use different approaches or don't call Probe() at all. None of them waits for the driver to be ready.  

@kubernetes-csi/csi-misc

### DIFF
--- a/connection/connection.go
+++ b/connection/connection.go
@@ -24,6 +24,9 @@ import (
 	"strings"
 	"time"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/kubernetes-csi/csi-lib-utils/protosanitizer"
 	"google.golang.org/grpc"
@@ -33,6 +36,9 @@ import (
 const (
 	// Interval of logging connection errors
 	connectionLoggingInterval = 10 * time.Second
+
+	// Interval of trying to call Probe() until it succeeds
+	probeInterval = 1 * time.Second
 )
 
 // Connect opens insecure gRPC connection to a CSI driver. Address must be either absolute path to UNIX domain socket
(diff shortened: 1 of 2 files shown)


connection/connection.go
```diff
@@ -177,3 +178,68 @@ func GetDriverName(ctx context.Context, conn *grpc.ClientConn) (string, error) {
...
+			continue
+		}
+		t := rpc.GetType()
+		caps[t] = true
+	}
+	return caps, nil
+}
+
+// Probe calls Probe() of a CSI driver and waits until the driver becomes ready.
+// TODO: wait until the driver is ready? It may require timeout instead of context as a parameter.
```
msau42: I think it makes sense to wait until the driver returns success.  Unsure if it needs to return on timeout or it can try forever like the connect

jsafrane: Reworked to try forever. It's expected that liveness probe will restart the driver if it takes too long.

jsafrane:
`Probe()` now tries with 1 second sleep until it succeeds. I'm not using exponential backoff because it would import  `kubernetes/apimachinery` and complicate dependent sidecars.

jsafrane:
Reworked a bit, `Probe()` now tries once (with given context) and should be used by livenessprobe, `ProbeForever()` tries forever and should be used by everything else.

msau42:
cc @andrewsykim

connection/connection.go
```diff
@@ -177,3 +184,111 @@ func GetDriverName(ctx context.Context, conn *grpc.ClientConn) (string, error) {
...
+		}
+		t := rpc.GetType()
+		caps[t] = true
+	}
+	return caps, nil
+}
+
+// ProbeForever calls Probe() of a CSI driver and waits until the driver becomes ready.
+// Any error other than timeout is returned.
+func ProbeForever(conn *grpc.ClientConn, singleProbeTimeout time.Duration) error {
```
msau42: does the singleProbeTimeout need to be configurable?

jsafrane: Some drivers may require longer timeout for their checks and only driver author knows that. Traditionally we use the same timeout as for the real CSI call (ControllerPublish, CreateVolume).

PR #18, opened 2019-02-14
Add option to exit on reconnect

jsafrane:
With optional creation of `/dev/termination-log`.

### DIFF
--- a/connection/connection.go
+++ b/connection/connection.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"strings"
 	"time"
@@ -34,6 +35,8 @@ const (
 	// Interval of logging connection errors
 	connectionLoggingInterval = 10 * time.Second
 )
+
+const terminationLogPath = "/dev/termination-log"
 
 // Connect opens insecure gRPC connection to a CSI driver. Address must be either absolute path to UNIX domain socket
 // file or have format '<protocol>://', following gRPC name resolution mechanism at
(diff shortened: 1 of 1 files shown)


connection/connection.go
```diff
@@ -65,12 +69,32 @@ type Option func(o *options)
...
+func ExitOnConnectionLoss(terminationLogPath string) Option {
+	return func(o *options) {
+		o.reconnect = func() bool {
+			terminationMsg := "Lost connection to CSI driver, exiting"
+			if terminationLogPath != "" {
+				if err := ioutil.WriteFile(terminationLogPath, []byte(terminationMsg), 0644); err != nil {
+					klog.Errorf("%s: %s", terminationLogPath, err)
+				}
+			}
+			klog.Fatalf(terminationMsg)
```
pohly: Won't that raise a panic? That might get caught further up in the call chain and even if it isn't.

What is the output? Does it include a stack backtrace? I don't think we want that - just the message and the exit the process.

jsafrane: It just prints the message, no panic.

```
I0214 16:59:44.965050   26907 shared_informer.go:123] caches populated
E0214 16:59:46.528572   26907 connection.go:127] Lost connection to unix:///tmp/csi.sock.
E0214 16:59:46.528679   26907 connection.go:85] /dev/termination-log: open /dev/termination-log: permission denied
F0214 16:59:46.528735   26907 connection.go:88] Lost connection to CSI driver, exiting

$ echo $?
255
```

connection/connection.go
```diff
@@ -65,12 +69,32 @@ type Option func(o *options)
 // connection got lost. If that callback returns true, the connection
 // is restablished. Otherwise the connection is left as it is and
 // all future gRPC calls using it will fail with status.Unavailable.
+// Only one of ExitOnConnectionLoss and OnConnectionLoss can be used.
 func OnConnectionLoss(reconnect func() bool) Option {
```
msau42: is anything going to use this option? do we still need it?

connection/connection.go
pohly: That's okay then. I was thinking of `klog.Panicf`, my bad.

connection/connection.go
pohly: The test code uses it. I think we should keep it.

jsafrane:
Reworked `ExitOnConnectionLoss` to return callback for `OnConnectionLoss`:

```
conn, err := connection.Connect(addr, OnConnectionLoss(ExitOnConnectionLoss()))
```

- it keeps OnConnectionLoss unit tests
- simplifies `ExitOnConnectionLoss` a bit (no unit tests there since it exits)

pohly:
lgtm, but I'll let @msau42 do the final approval because she also might have an opinion.

PR #19, opened 2019-02-15
add optional timeout to CSI lib connection utils

andrewsykim:
Allows users of the connection lib to set an optional timeout. If no timeout is set, the default is to retry forever. 

/assign @msau42 @jsafrane @vladimirvivien

### DIFF
--- a/connection/connection.go
+++ b/connection/connection.go
@@ -58,7 +58,7 @@ const terminationLogPath = "/dev/termination-log"
 // For other connections, the default behavior from gRPC is used and
 // loss of connection is not detected reliably.
 func Connect(address string, options ...Option) (*grpc.ClientConn, error) {
-	return connect(address, []grpc.DialOption{}, options)
+	return connect(address, options)
 }
 
 // Option is the type of all optional parameters for Connect.
@@ -71,6 +71,14 @@ type Option func(o *options)
 func OnConnectionLoss(reconnect func() bool) Option {
 	return func(o *options) {
 		o.reconnect = reconnect
+	}
+}
+
+// WithTimeout sets the connection timeout duration
+// If not specified, all connections will retry forever.
+func WithTimeout(timeout time.Duration) Option {
+	return func(o *options) {
+		o.timeout = timeout
 	}
 }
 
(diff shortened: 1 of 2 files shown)


connection/connection.go
```diff
@@ -89,21 +97,28 @@ func ExitOnConnectionLoss() func() bool {
 
 type options struct {
 	reconnect func() bool
+	timeout   time.Duration
 }
 
 // connect is the internal implementation of Connect. It has more options to enable testing.
-func connect(address string, dialOptions []grpc.DialOption, connectOptions []Option) (*grpc.ClientConn, error) {
```
andrewsykim: Decided to remove this because I don't think it makes sense to pass in dial options into an unexported function.

connection/connection.go
```diff
@@ -74,6 +74,14 @@ func OnConnectionLoss(reconnect func() bool) Option {
 	}
 }
 
+// WithTimeout sets the connection timeout duration
+// If not specified, all connections will retry forever.
+func WithTimeout(timeout time.Duration) Option {
```
pohly: A more flexible solution for specifying a timeout is to add a `context.Context` parameter to the `Connect` function. Then one can specify both a deadline and a timeout, depending on the situation. If we expose a timeout option at all in the API, then I would prefer the solution based on context.

@andrewsykim why do you think such an option is needed?

When @jsafrane added this function, he intentionally limited the API to just what the CSI sidecars need, to ensure consistent behavior. That's why the testing is partly using an internal helper function.

msau42: The connect forever option doesn't work well with the liveness probe sidecar because the health check will hang instead of returning an error.

@andrewsykim can you confirm if the Kubernetes liveness probe can appropriately handle the health check hanging?

andrewsykim: IIRC, the default (and minimum) livenessprobe timeout on kubelet is 1s, so liveness check hanging and not returning an error would still allow the kubelet to restart the container within the timeout specified on the container spec.

If we're okay assuming this, then maybe a timeout option is not needed at all. If we do want the livenessprobe sidecar to explicitly return 500 when a timeout is reached then being able to pass in a timeout via the connection lib would be useful.

andrewsykim: > A more flexible solution for specifying a timeout is to add a `context.Context` parameter to the `Connect` function. Then one can specify both a deadline and a timeout, depending on the situation. If we expose a timeout option at all in the API, then I would prefer the solution based on context.

I can update to use context instead, though we may lose some of the functionality we get passing in the timeout as a grpc dial option. Either works for me.

msau42: > liveness check hanging and not returning an error would still allow the kubelet to restart the container within the timeout specified

Then I think in that case, it is fine to have the liveness probe hang on connect(). The only downside I see, is that healthz requests may pile up on the liveness probe if the driver is forever in an unready state.

How about something like this for the liveness probe:
1. Instead of connecting to the driver on every healthz check, connect to it once in the beginning. This can try to connect forever.
2. Whenever a healthz request comes, call Probe() with timeout, so that we can return a 500 immediately and not have outstanding requests.

pohly: What functionality could get lost?

Note that gRPC already deprecated `WithTimeout` in favor of `context.WithTimeout` - https://godoc.org/google.golang.org/grpc#WithTimeout.

andrewsykim: > How about something like this for the liveness probe:
1. Instead of connecting to the driver on every healthz check, connect to it once in the beginning. This can try to connect forever.
2. Whenever a healthz request comes, call Probe() with timeout, so that we can return a 500 immediately and not have outstanding requests.

Sounds good. I'll update livenessprobe to do this. 

> Note that gRPC already deprecated `WithTimeout` in favor of `context.WithTimeout` - https://godoc.org/google.golang.org/grpc#WithTimeout.

I didn't know this, this is good to know thanks!

andrewsykim:
/close

Closing for now, see https://github.com/kubernetes-csi/csi-lib-utils/pull/19#discussion_r258213629 for more details

PR #20, opened 2019-02-25
Move common CSI driver RPCs from package connection -> rpc

andrewsykim:
Before updating a bunch of CSI sidecars to use the common CSI RPCs `GetDriverName`, `GetPluginCapabilities`, `GetControllerCapabilities` and `Probe`, I wanted to propose that we move these to its own package called `rpc`. This keeps the `connection` package cleaner and only hold code related to connection management. 

/assign @msau42 @jsafrane


jsafrane:
Do we care about backward compatibility? It will break anyone who uses this library. 

Otherwise LGTM.

rpc/common_test.go
```diff
@@ -0,0 +1,611 @@
...
+	probeCalls     []probeCall
+	probeCallCount int
+}
+
+type probeCall struct {
+	response *csi.ProbeResponse
+	err      error
+}
+
+var _ csi.IdentityServer = &identityServer{}
```
msau42: nit: can you prepend "fake" to the name

rpc/common_test.go
```diff
@@ -0,0 +1,611 @@
...
+	i.probeCallCount++
+	return call.response, call.err
+}
+
+type controllerServer struct {
+	controllerGetCapabilitiesResponse *csi.ControllerGetCapabilitiesResponse
+	err                               error
+}
+
+var _ csi.ControllerServer = &controllerServer{}
```
msau42: nit: can you prepend "fake" to the name

rpc/common_test.go
andrewsykim: done!

rpc/common_test.go
andrewsykim: done!

andrewsykim:
Thanks @msau42! Can you let me know when a new release is cut with this change?

msau42:
https://github.com/kubernetes-csi/csi-lib-utils/releases/tag/v0.4.0-rc1

Issue #15, opened 2019-02-08
Common leader election library


msau42:
Let's standardize on using the [Lease](https://github.com/kubernetes/kubernetes/pull/70778) mechanism for all the sidecars, and try to have common arguments too.

cc @wongma7 @xing-yang

jsafrane:
I am 100% for standardizing leader election, there is just one glitch: it will pull lot of Kubernetes dependencies into this library. It will complicate rebases and releases - with new Kubernetes, one needs to update csi-lib-utils and external-provisioner-lib, wait until they release a new version and then bump their versions in all sidecars that use it. Is there a way how to do it better / faster?

andrewsykim:
> It will complicate rebases and releases - with new Kubernetes, one needs to update csi-lib-utils and external-provisioner-lib, wait until they release a new version and then bump their versions in all sidecars that use it. Is there a way how to do it better / faster?

Can we go back to mono repos? 😆

xing-yang:
Just read Jan's comments.  I see what he meant.  Right now this lib does not have dependency on kubernetes.  Once we add this function in here, this lib will have dependencies on kubernetes.  We need to sync up kubernetes versions all the way up to all sidecar containers that use this lib.  I don't know if there's a better way.  One alternative is to leave it to each sidecar container to implement this, but then we don't have a common lib to use.

msau42:
/close
Done in #21 

We can hopefully have better dependency management with go modules.

PR #21, opened 2019-03-04
Add common leader election library

andrewsykim:
Adds a common leader election library that will be used by CSI sidecars. Currently supports leader election with resource type `Endpoints` and `ConfigMaps`. I wanted to get early feedback on this PR so I left out `Lease` for now, as soon as a beta version of v1.14 is released in `k8s.io/client-go`, I will update the PR to support `Lease` based leader election. 

https://github.com/kubernetes-csi/csi-lib-utils/issues/15

/assign @msau42 @xing-yang


leaderelection/leader_election.go
```diff
@@ -0,0 +1,165 @@
...
+	resourceLock string
+
+	leaseDuration time.Duration
+	renewDeadline time.Duration
+	retryPeriod   time.Duration
+
+	clientset kubernetes.Interface
+}
+
+func NewConfigMapLeaderElection(clientset kubernetes.Interface, lockName, lockNamespace string, runFunc func(ctx context.Context)) *leaderElection {
```
msau42: I think we only want to support Lease going forward and not ConfigMap nor Endpoints

andrewsykim: I left this here for backwards compat because external attacher still uses ConfigMap and external provisioner still uses Endpoints. I updated the PR so that `NewLeaderElection` defaults to Lease type, I can remove it otherwise :)

andrewsykim:
Update: I ended up vendoring in `kubernetes-v1.14.0-alpha.3` so I updated the PR to enable lease based leader election. We probably shouldn't merge this until there's at least a beta version of v1.14 though.

leaderelection/leader_election.go
msau42: I'm not sure how much benefit we get from still having the option to support the other mechanisms. It means the sidecar rbacs will still be different and may behave differently (and have different bugs) depending on the mechanism.

Another interesting side-effect that I'm not sure how widespread it is, but if the CSI driver implements some sort of in-memory caching of state that is shared between provisioner/attach/resize/snapshot/etc, then the fact that different sidecars from different pods could be the active leader could cause inconsistencies. (ie provision called in one pod, attach called from another pod).  I guess for that case, we just need to say that if you support running your driver with leader election, then it needs to be resilient to this scenario.

jsafrane: I'd like to use Lease everywhere and remove ConfigMap / Endpoints, if it is possible. Do we need to be backward compatible in this case? IMO the only change is RBAC.

msau42: The risk is that you may have two leaders during a rolling update of the driver because some pods will use the old type, and the new ones will use Lease.

What is the consequence of having two leaders?  You may provision/delete/attach/detach twice?  I think provision/delete twice is safe to do, but I'm not sure about attach/detach twice.  Can all control planes handle it?

jsafrane:
Btw, vendoring any pre-release Kubernetes API will be interesting, because various components (at least external snapshotter & provisioner) need to agree on single version that contains changes for all of them.

leaderelection/leader_election.go
andrewsykim: PR updated to only support leases

msau42: I'm still concerned that we could be breaking some driver for attach/detach that we don't know about.  Given that we never called out leader-election as an alpha/beta feature, I think we should play it on the safe side and deprecate the old way while introducing the option to use the new Lease.  Maybe it's worth having all the methods in the library then so at least the sidecar code is not super ugly.

andrewsykim: Sounds good to me, I'll update the PR!

andrewsykim:
This is ready for review, imported kubernetes v1.14-beta.2 for now

PR #22, opened 2019-03-19
deprecated flag handling

pohly:
Several sidecars (for example, external-provisioner) deprecate command
line flags. When that happens, the flag must still be treated the same
way as before (i.e. optional values for boolean, required values for
non-boolean), but trying to use the flag needs to trigger a
deprecation warning.

An alternative to this code is to let the flag set a variable and then
check the variable. But that only catches cases where the variable is
set to a non-default value and leaves the variable in the code, which
can lead to mistakes (see
https://github.com/kubernetes-csi/external-provisioner/pull/239).

### DIFF
--- a/deprecatedflags/deprecatedflags.go
+++ b/deprecatedflags/deprecatedflags.go
@@ -1,0 +1,57 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


pohly:
For usage of this package, see https://github.com/kubernetes-csi/external-provisioner/pull/255.

That PR can either be merged and later the code gets replaced, or we merge this one first, do a release, and then update the other PR.

/cc @jsafrane

jsafrane:
/lgtm

Thanks!

PR #24, opened 2019-04-01
discover in cluster namespace for leader election

andrewsykim:
Signed-off-by: Andrew Sy Kim <kiman@vmware.com>

Adds support to discover namespaces in cluster by checking the env var POD_NAMESPACE, then the file `/var/run/secrets/kubernetes.io/serviceaccount/namespace` (should exist in every pod). Falls back to default.

### DIFF
--- a/leaderelection/leader_election.go
+++ b/leaderelection/leader_election.go
@@ -19,8 +19,10 @@ package leaderelection
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"regexp"
+	"strings"
 	"time"
 
 	"k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


leaderelection/leader_election.go
```diff
@@ -185,3 +197,20 @@ func sanitizeName(name string) string {
 	}
 	return name
 }
+
+// inClusterNamespace returns the namespace in which the pod is running in by checking
+// the env var POD_NAMESPACE, then the file /var/run/secrets/kubernetes.io/serviceaccount/namespace.
+// if neither returns a valid namespace, the "default" namespace is returned
+func inClusterNamespace() string {
```
andrewsykim: I prefer not to fall back to `default` here and just return an error. I left the fallback to default for now since this is used in external-provisioner. Do we know of any situations where a sidecar wouldn't run in the cluster? If not I think the fallback to `default` should be removed and instead exit with error

msau42: Can we say that for Lease mechanism (which should be all new), require a namespace field to be set if not found, but for the other two, we allow the default for backwards compatibility?

andrewsykim: We could, though the implementation will be a bit messy I think. I added the `WithNamespace` method for the cases where we need to account for backwards compatibility. I guess the this boils down to two questions:
1) how reliable is it for the namespace file to exist (I assume this is not going to change anytime in the near future)
2) would any user run this "out of cluster"

msau42: 1. I don't expect this to change. It would probably be considered an API break
2. It's possible someone may want to run this out of cluster, but we can wait for a feature request to add the option back.

andrewsykim: I'm okay with the current PR merging as is, but having the `--leader-election-namespace` flag as optional in all the sidecars. Setting `--leader-election-namespace` calls the namespace override with `WithNamespace`. Thoughts?

msau42: I think it's fine to maintain backwards compatibility.  If the default doesn't work, then the user needs to explicitly specify something.

PR #26, opened 2019-04-08
Prow testing

pohly:
This enables unit testing. It would be possible to write a custom E2E test for leader election, but that can be added later.

/assign @msau42

### DIFF
--- a/rpc/common_test.go
+++ b/rpc/common_test.go
@@ -241,6 +241,9 @@ func TestGetPluginCapabilities(t *testing.T) {
 			defer os.RemoveAll(tmp)
 			identity := &fakeIdentityServer{
 				getPluginCapabilitiesResponse: test.output,
+
+				// Make code compatible with gofmt 1.10.2 (used by pull-sig-storage-csi-lib-utils-stable)
+				// and 1.11.1 (which will be used by new Prow job) via an extra blank line.
 				err: injectedErr,
 			}
 			addr, stopServer := startServer(t, tmp, identity, nil)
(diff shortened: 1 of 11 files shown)


pohly:
/hold

This tentatively uses https://github.com/kubernetes-csi/csi-release-tools/pull/13, which needs to be merged first. We also need https://github.com/kubernetes/test-infra/pull/12088 before this can be tested.

rpc/common_test.go
```diff
@@ -241,7 +241,7 @@ func TestGetPluginCapabilities(t *testing.T) {
 			defer os.RemoveAll(tmp)
 			identity := &fakeIdentityServer{
 				getPluginCapabilitiesResponse: test.output,
-				err: injectedErr,
+				err:                           injectedErr,
```
pohly: I'm not sure why this passed testing in https://github.com/kubernetes-csi/csi-lib-utils/pull/20.

pohly: The explanation is that the Docker image that is currently used by the mandatory pre-merge check l only has Go 1.10.1. We'll transition to current Go as part of replacing the Prow check.

pohly:
@msau42 are we ready to replace pull-sig-storage-csi-lib-utils-stable (the custom Prow job) with pull-sig-storage-csi-lib-utils (the generated one) and enable also mandatory unit testing for csi-release-tools and csi-test?

https://github.com/kubernetes/test-infra/blob/24aefcb76ce14366c3e4380aebcb1ec7eb3664e7/config/jobs/kubernetes-csi/gen-jobs.sh#L310-L328

I think we can and should. I'd also like to rename to `pull-kubernetes-csi-<repo>`, for the sake of consistency.

msau42:
Sounds fine? Which prs need to go in first?

msau42:
Did you want to rename the jobs first?

PR #32, opened 2019-09-24
update to k/k 1.16

jichenjc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

solving https://github.com/kubernetes/cloud-provider-openstack/issues/702 as I have to include leaderelection file into cloud provider openstack and when running `dep ensure` I have following error, seems csi-lib-utils still rely on k/k1.14 while conflict to cloud provider openstack on 1.16

Solving failure: No versions of github.com/kubernetes-csi/csi-lib-utils met constraints:
        v0.6.1: Could not introduce github.com/kubernetes-csi/csi-lib-utils (from https://github.com/kubernetes-csi/csi-lib-utils)@v0.6.1, as it has a dependency on k8s.io/client-go with constraint kubernetes-1.14.0, which has no overlap with existing constraint kubernetes-1.16.0 from (root)

Fixes #

```release-note
NONE
```


pohly:
/assign @andrewsykim

Can you have a look at this? I don't know what the implications will be for other users of this lib if we bump the Kubernetes dependency to 1.16.

The underlying problem is that Kubernetes packages don't have semantic versioning. So even though csi-lib-utils seems to be compatible with any version >= 1.14 (or at least no code changes were necessary to compile with 1.16), we can't specify that in the `Gopkg.toml` file.

I don't know for sure how Go modules would handle this. I suspect it would allow us to select 1.14 for reproducible builds of `csi-lib-utils`, but as this isn't a minimal viable version with semantic versioning, users might accidentally pick something older.

pohly:
Perhaps we can approximate the `go mod` behavior by removing the constraints from `Gopkg.toml`?

jichenjc:
can we create a new tag e.g 0.6.2 if we bump to k/k 1.16 so older still can rely on 0.6.1? and for newer can rely on 0.6.2

the go mod might be another way to try as well

andrewsykim:
@jichenjc would an override in your Gopkg.toml file work for the kubernetes dependencies work?

pohly:
@jichenjc: can you try whether the switch to go modules (https://github.com/kubernetes-csi/csi-lib-utils/pull/33) works for you?

jichenjc:
ok, will try it later on (holiday approaching), thanks~

pohly:
PR #33 was merged, please let us know whether this is still needed.

jichenjc:
let me close this and let me try with go module first, if needed, might open another one. thanks~

PR #33, opened 2019-09-26
Switch to go modules

pohly:
**What type of PR is this?**
/kind cleanup

We investigated the move from "dep" to "go mod" for a while already, see https://github.com/kubernetes-csi/csi-test/pull/188. `csi-lib-utils` is a good starting point for that because it only has a few dependencies, none of them in `kubernetes-csi`.

A reason for doing the transition now is the problem mentioned in https://github.com/kubernetes-csi/csi-lib-utils/pull/32: the tag names in `Gopk.toml` prevent consumers from using a newer Kubernetes, even though the `csi-lib-utils` code itself seems to be compatible with it.

I verified that `external-provisioner` can still vendor the updated `csi-lib-utils` with `dep`, see https://github.com/pohly/external-provisioner/commits/csi-lib-utils-vendoring. That same branch also shows that updating to Kubernetes 1.16 is possible.

```release-note
vendoring and dependencies managed with "go mod"
```


go.mod
```diff
@@ -0,0 +1,34 @@
+module github.com/kubernetes-csi/csi-lib-utils
+
+go 1.12
+
+require (
```
pohly: In contrast to the manually written `Gopk.toml`, `go.mod` is complete and lists all dependencies with revisions chosen at the time of running `go mod init`. We have less control over dependency resolution (can't specify the exact minimum version that the code is compatible with, can't specify ranges), but that's simply how "go mod" works. :man_shrugging:

pohly: A diff of `vendor` shows that we still use the same source code. Just pruning is handled differently, so some files are gone and others are new.

Compilation is done without Go modules, so nothing changes when building.

msau42:
Thanks! Can you also add something to the release-tools/README about what commands you used, and also how to update in the future.

/lgtm

pohly:
> Can you also add something to the release-tools/README about what commands you used, and also how to update in the future.

Done, see https://github.com/kubernetes-csi/csi-release-tools/pull/34

While working on that I noticed that I missed a `go mod tidy` invocation in this repo. I added a check for that in that PR.

PR #34, opened 2019-10-04
clean up go.mod

pohly:
**What type of PR is this?**
/kind cleanup

The current `go.mod` file lacks some indirect dependencies. `go mod tidy` adds those.

We want to detect this in the future, which is covered by the extended vendor check.

```release-note
NONE
```


release-tools/README.md
```diff
@@ -106,3 +106,37 @@ Kubernetes releases:
...
+
+`go mod tidy` must be used to ensure that the listed dependencies are
+really still needed. Changing import statements or a tentative `go
+get` can result in stale dependencies.
+
+The `test-vendor` verifies that it was used when run locally or in a
+pre-merge CI job. If a `vendor` directory is present, it will also
+verify that it's content is up-to-date.
+
+The `vendor` directory is optional. It is still present in projects
```
msau42: I remember hearing from sig-testing that prow jobs have some sort of go mode cache available. We can look into using it as a next step after converting our repos to go modules

pohly: I think the conclusion was to simply use the official Go mirror, something that 1.13 does automatically (https://golang.org/doc/go1.13#introduction) and was enabled for 1.12 via an env variable.

We can convert with `vendor` and then check whether it affects build times when we drop it.

pohly:
@msau42: `csi-release-tools` updated to latest master, ready for merging.

PR #35, opened 2019-12-23
Introduce a CSI Metrics Library

saad-ali:
**What type of PR is this?**
/kind feature

This PR adds a library that automatically records prometheus metrics for every CSI gRPC call.

Fixes #

```release-note
Introduce a CSI Metrics Library that can be used to automatically record prometheus metrics for every CSI gRPC call.
```

/assign @msau42 
/assign @verult


saad-ali:
Not sure why `test-vendor` is failing. `GO111MODULE=on go mod tidy` and `GO111MODULE=on go mod vendor` return not changes for me.

pohly:
> Not sure why test-vendor is failing. GO111MODULE=on go mod tidy and GO111MODULE=on go mod vendor return no changes for me. Suggestion to fix welcome.

Are you using the same Go version as this repo? The content of the `vendor` directory is sensitive to that.

We added a check and a helper script to deal with this (see https://github.com/kubernetes-csi/csi-release-tools/pull/41 and https://groups.google.com/forum/#!topic/kubernetes-sig-storage-wg-csi/LBTRf1AqRXA), but this repo still has an older version of the csi-release-tools scripts.

pohly:
> Are you using the same Go version as this repo?

Hmmm, I need to add this repo to my list of repos that need to be updated when changing csi-release-tools - this repo still specifies Go 1.12.4 as the reference version (https://github.com/kubernetes-csi/csi-lib-utils/blob/3db811614bd14d294eb08069b7ac59bd46f69139/release-tools/travis.yml#L7).

Let's merge https://github.com/kubernetes-csi/csi-lib-utils/pull/36 first.

metrics/metrics.go
```diff
@@ -0,0 +1,161 @@
...
+
+const (
+	// Common metric strings
+	subsystem             = "csi_sidecar"
+	labelCSIDriverName    = "driver_name"
+	labelCSIOperationName = "method_name"
+	labelGrpcStatusCode   = "grpc_status_code"
+	unknownCSIDriverName  = "unknown-driver"
+
+	// CSI Operation Total with status code - Count Metric
```
msau42: Do we still need this? xref: https://github.com/kubernetes-csi/external-provisioner/pull/386#discussion_r359129206

saad-ali: Ack, will remove.

connection/connection.go
```diff
@@ -58,8 +59,8 @@ const terminationLogPath = "/dev/termination-log"
 //
 // For other connections, the default behavior from gRPC is used and
 // loss of connection is not detected reliably.
-func Connect(address string, options ...Option) (*grpc.ClientConn, error) {
-	return connect(address, []grpc.DialOption{}, options)
+func Connect(address string, metricsManager metrics.CSIMetricsManager, options ...Option) (*grpc.ClientConn, error) {
```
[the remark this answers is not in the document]
saad-ali: Yes. But I'll add a helper function here to make that easier.

metrics/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+		err := http.ListenAndServe(metricsAddress, nil)
+		if err != nil {
+			klog.Fatalf("Failed to start prometheus metrics endpoint on specified address (%q) and path (%q): %s", metricsAddress, metricsPath, err)
+		}
+	}()
+}
+
+// VerifyMetricsMatch is a helper function that verifies that the expected and
+// actual metrics are identical excluding metricToIgnore.
+func VerifyMetricsMatch(expectedMetrics, actualMetrics string, metricToIgnore string) error {
```
msau42: This is only used in test right? Can we move it to the test file instead of exporting it?

saad-ali: This method is only used by tests. Ideally it should be in the _test file,  but *_test.go files are compiled into the package only when running go test for that package and this method is used by metrics_test as well as connection_test. In the future, if there are more consumers, we can consider moving it to a new, standalone package. For now, I added a comment documenting this.

verult:
@mattcary @saikat-royc

PR #44, opened 2020-06-18
leaderelection add context

aimuz:
**What type of PR is this?**
/kind feature

Add context

```release-note
leaderelection add WithContext func
```

### DIFF
--- a/leaderelection/leader_election.go
+++ b/leaderelection/leader_election.go
@@ -60,6 +60,8 @@ type leaderElection struct {
 	renewDeadline time.Duration
 	retryPeriod   time.Duration
 
+	ctx context.Context
+
 	clientset kubernetes.Interface
 }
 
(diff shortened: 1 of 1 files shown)


leaderelection/leader_election.go
```diff
@@ -174,7 +181,10 @@ func (l *leaderElection) Run() error {
 		},
 	}
 
-	leaderelection.RunOrDie(context.TODO(), leaderConfig)
+	if l.ctx == nil {
```
pohly, line 184: Nit: I don't like that it modifies `l` here. I can't quite nail it, but it feels wrong.

Also, why "context.TODO()"? That implies that the code needs further work. What is that missing work? If this is final, use `context.Background()'. If not, explain what needs to be done.

I'd prefer:
```
ctx := l.ctx
if ctx == nil {
    ctx = context.Background()
}
leaderelection.RunOrDie(ctx, leaderConfig)
```

aimuz: use "context.TODO()" because it was like this before, but now it seems to be true "context.Background()" better.

I have recommitted the code

aimuz:
@saad-ali

PR #49, opened 2020-08-20
metrics: support usage inside CSI driver

pohly:
**What type of PR is this?**
/kind feature

A CSI driver is a gRPC server, which implies that the interceptor must
use a slightly different API. The extended CSIMetricsManager gets
exported with a suitable method for that.

To avoid counting the same operation twice in the same metric, CSI
sidecar and driver should use different subsystem names. CSI driver
authors may decide to declare metrics support as "stable" based on
other criteria than the Kubernetes-CSI sidecars. Both aspects are now
configurable.

These API changes are backwards-compatible.

```release-note
metrics support can also be used by CSI drivers
```

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -229,7 +229,7 @@ func (cmm *csiMetricsManager) RecordMetrics(
 	operationName string,
 	operationErr error,
 	operationDuration time.Duration) {
-	cmm.recordMetricsWithLabels(operationName, operationErr, operationDuration)
+	cmm.recordMetricsWithLabels(operationName, operationErr, operationDuration, nil)
 }
 
 // recordMetricsWithLabels is the internal implementation of RecordMetrics.
@@ -237,18 +237,10 @@ func (cmm *csiMetricsManager) recordMetricsWithLabels(
 	operationName string,
 	operationErr error,
 	operationDuration time.Duration,
-	labelValues ...string) {
+	labelValues map[string]string) {
 	values := []string{cmm.driverName, operationName, getErrorCode(operationErr)}
-	toAdd := len(labelValues)
-	if toAdd > len(cmm.additionalLabelNames) {
-		// To many labels?! Truncate. Shouldn't happen because of
-		// error checking in WithLabelValues.
-		toAdd = len(cmm.additionalLabelNames)
-	}
-	values = append(values, labelValues[0:toAdd]...)
-	for i := toAdd; i < len(cmm.additionalLabelNames); i++ {
-		// Backfill missing values with empty string.
-		values = append(values, "")
+	for _, name := range cmm.additionalLabelNames {
+		values = append(values, labelValues[name])
 	}
 	for _, label := range cmm.additionalLabels {
 		values = append(values, label.value)
@@ -260,34 +252,42 @@ type csiMetricsManagerWithValues struct {
 	*csiMetricsManager
 
 	// additionalValues holds the values passed via WithLabelValues.
-	additionalValues []string
+	additionalValues map[string]string
 }
 
 // WithLabelValues in the base metrics manager creates a fresh wrapper with no labels and let's
 // that deal with adding the label values.
 func (cmm *csiMetricsManager) WithLabelValues(labels map[string]string) (CSIMetricsManager, error) {
-	cmmv := &csiMetricsManagerWithValues{csiMetricsManager: cmm}
+	cmmv := &csiMetricsManagerWithValues{
+		csiMetricsManager: cmm,
+		additionalValues:  map[string]string{},
+	}
 	return cmmv.WithLabelValues(labels)
 }
 
 // WithLabelValues in the wrapper creates a wrapper which has all existing labels and
-// adds the new ones, with error checking.
+// adds the new ones, with error checking. Can be called multiple times. Each call then
+// can add some new value(s). It is an error to overwrite an already set value.
+// If RecordMetrics is called before setting all additional values, the missing ones will
+// be empty.
 func (cmmv *csiMetricsManagerWithValues) WithLabelValues(labels map[string]string) (CSIMetricsManager, error) {
-	extended := &csiMetricsManagerWithValues{cmmv.csiMetricsManager, append([]string{}, cmmv.additionalValues...)}
-
-	for name := range labels {
-		if !cmmv.haveAdditionalLabel(name) {
+	extended := &csiMetricsManagerWithValues{
+		csiMetricsManager: cmmv.csiMetricsManager,
+		additionalValues:  map[string]string{},
+	}
+	// We need to copy the old values to avoid modifying the map in cmmv.
+	for name, value := range cmmv.additionalValues {
+		extended.additionalValues[name] = value
+	}
+	// Now add all new values.
+	for name, value := range labels {
+		if !extended.haveAdditionalLabel(name) {
 			return nil, fmt.Errorf("label %q was not defined via WithLabelNames", name)
 		}
-	}
-	// Add in same order as in the label definition.
-	for _, name := range cmmv.additionalLabelNames {
-		if value, ok := labels[name]; ok {
-			if len(cmmv.additionalValues) >= len(extended.additionalLabelNames) {
-				return nil, fmt.Errorf("label %q = %q cannot be added, all labels already have values %v", name, value, cmmv.additionalValues)
-			}
-			extended.additionalValues = append(extended.additionalValues, value)
+		if v, ok := extended.additionalValues[name]; ok {
+			return nil, fmt.Errorf("label %q already has value %q", name, v)
 		}
+		extended.additionalValues[name] = value
 	}
 	return extended, nil
 }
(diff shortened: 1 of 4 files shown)


pohly:
/assign @msau42 

Can you perhaps review? This is based on what we discussed on Slack, so you have some context here.

metrics/metrics.go
```diff
@@ -73,25 +80,47 @@ type CSIMetricsManager interface {
 	StartMetricsEndpoint(metricsAddress, metricsPath string)
 }
 
+// MetricsManagerOption is used to pass optional configuration to a
+// new metrics manager.
+type MetricsManagerOption func(*csiMetricsManager)
+
+// WithSubsystem overrides the default subsystem name.
+func WithSubsystem(subsystem string) MetricsManagerOption {
```
msau42, line 88: Do we want the subsystem to be configurable or should we just have 2 constructor APIs, one for sidecars and one for plugins?

What happens if someone decides to input their own subsystem name?

metrics/metrics.go
```diff
@@ -73,25 +80,47 @@ type CSIMetricsManager interface {
...
+	return func(cmm *csiMetricsManager) {
+		cmm.subsystem = subsystem
+	}
+}
+
+// WithStabilityLevel overrides the default stability level.
+func WithStabilityLevel(stabilityLevel metrics.StabilityLevel) MetricsManagerOption {
```
msau42, line 95: Do we want stability level to be configurable? If someone sets it to stable, what happens if the library decides to change the schema?

metrics/metrics.go
pohly: > Do we want the subsystem to be configurable or should we just have 2 constructor APIs, one for sidecars and one for plugins?

Configurable, please. In PMEM-CSI, CreateVolume is passed from sidecar to controller to a driver on a node. All three have metrics gathering enabled. The same argument about counting the same operation multiple times also applies there, i.e. all three must use different subsystem. I ended up using "csi_sidecar", "csi_controller", "csi_node". 

> What happens if someone decides to input their own subsystem name?

My assumption was that a CSI driver is free to choose the names for itself. But I am not sure anymore, because the names are the same regardless of their origin. But this is general issue with Prometheus, right? One component can provide a metric "foobar" with a certain semantic and another component some other metric "foobar" with a completely different semantic.

metrics/metrics.go
pohly: > Do we want stability level to be configurable? If someone sets it to stable, what happens if the library decides to change the schema?

The user cannot use that modified library and has to fork it if it wants to continue supporting the old values. Clearly overriding "alpha" with "stable" is a bad idea for that reason.

I was more interested in the other direction: a downstream user might enable metrics support as an alpha feature and may want to keep the alpha label even when the library moves to "stable".

metrics/metrics.go
msau42: Can we have both? Standard constructors for the 2 most common use cases: a sidecar or a plugin.  And a 3rd customizable option if the driver is doing some advanced stuff.

Standardizing on a component name for plugins would be really nice so we can write generic collectors that work for all csi drivers.

pohly: > Can we have both? Standard constructors for the 2 most common use cases: a sidecar or a plugin. And a 3rd customizable option if the driver is doing some advanced stuff.

Okay.

> Standardizing on a component name for plugins would be really nice so we can write generic collectors that work for all csi drivers.

With collectors, do you mean Prometheus jobs for scraping? Those don't depend on the metrics names. More problematic for scraping is how pods are labeled: there's no established convention for scraping multiple containers in a pod. I made a proposal in https://github.com/helm/charts/issues/22899 but got no response.

Standardizing the names becomes more interesting when running queries or creating dashboards.

msau42: Yes sorry I meant dashboards.

metrics/metrics.go
```diff
@@ -115,12 +201,18 @@ func (cmm *csiMetricsManager) GetRegistry() metrics.KubeRegistry {
...
 // operationDuration - time it took for the operation to complete
+// labelValues - for each additional label that was defined in WithLabels(), one value has to be passed (usually none)
 func (cmm *csiMetricsManager) RecordMetrics(
 	operationName string,
 	operationErr error,
-	operationDuration time.Duration) {
-	cmm.csiOperationsLatencyMetric.WithLabelValues(
-		cmm.driverName, operationName, getErrorCode(operationErr)).Observe(operationDuration.Seconds())
+	operationDuration time.Duration,
+	labelValues ...string) {
```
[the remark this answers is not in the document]
pohly: > So the order of the labels has to be in the same order as the keys?

Yes.

> Would it be better to have a map passed in and order by keys? Should we validate that the length of the values is the same as the length of the keys?

We could make the call more robust against incorrect usage; both of this would work. Map corresponds to the `GetMetricWith` method (https://godoc.org/github.com/prometheus/client_golang/prometheus#CounterVec.GetMetricWith), checking the length to `With` (https://godoc.org/github.com/prometheus/client_golang/prometheus#CounterVec.With). Either way, we then also should introduce an error return value and expect that the callers handle this.

I don't have a strong preference either way.

metrics/metrics.go
pohly: I think a note is enough. If someone really wants to increase stability, they have been warned and will have to live with the consequences.

metrics/metrics.go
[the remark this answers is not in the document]
pohly: One further comment: `labelValues ...string` was also chosen because it's compatible with current code which doesn't pass any additional labels. We could introduce `WithLabels` and/or `WithLabelValues` methods that return a CSIMetricsManager which uses the additional labels when its RecordMetrics is invoked. RecordMetrics itself then has the same parameters as it has before this PR.

This is probably the best solution, but also the most complicated one. I'm wondering whether it is overkill.

@msau42 do you have any preferences?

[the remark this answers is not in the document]
msau42: I think it's less brittle to not depend on the ordering of the values to match what was configured. If we need to change the interface to return an error, then I would return a new method for those drivers that want to use this customization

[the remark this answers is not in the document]
pohly: I revised the API such that `RecordMetrics` is as before and additional labels get injects as in Prometheus with a wrapper. I also added tests.

metrics/metrics.go
pohly: Added.

metrics/metrics.go
```diff
@@ -101,26 +208,105 @@ var _ CSIMetricsManager = &csiMetricsManager{}
...
+	return cmmv.WithLabelValues(labels)
+}
+
+// WithLabelValues in the wrapper creates a wrapper which has all existing labels and
+// adds the new ones, with error checking.
+func (cmmv *csiMetricsManagerWithValues) WithLabelValues(labels map[string]string) (CSIMetricsManager, error) {
+	extended := &csiMetricsManagerWithValues{cmmv.csiMetricsManager, append([]string{}, cmmv.additionalValues...)}
+
+	for name := range labels {
+		if !cmmv.haveAdditionalLabel(name) {
```
msau42, line 279: Is this needed or can you just verify the lengths at the end (and keep the duplicate values check)

metrics/metrics.go
```diff
@@ -101,26 +208,105 @@ var _ CSIMetricsManager = &csiMetricsManager{}
...
+	values := []string{cmm.driverName, operationName, getErrorCode(operationErr)}
+	toAdd := len(labelValues)
+	if toAdd > len(cmm.additionalLabelNames) {
+		// To many labels?! Truncate. Shouldn't happen because of
+		// error checking in WithLabelValues.
+		toAdd = len(cmm.additionalLabelNames)
+	}
+	values = append(values, labelValues[0:toAdd]...)
+	for i := toAdd; i < len(cmm.additionalLabelNames); i++ {
+		// Backfill missing values with empty string.
```
msau42, line 250: I'm not sure if we can make the assumption that the missing labels are all at the end.

metrics/metrics.go
```diff
@@ -101,26 +208,105 @@ var _ CSIMetricsManager = &csiMetricsManager{}
...
+		}
+	}
+	// Add in same order as in the label definition.
+	for _, name := range cmmv.additionalLabelNames {
+		if value, ok := labels[name]; ok {
+			if len(cmmv.additionalValues) >= len(extended.additionalLabelNames) {
+				return nil, fmt.Errorf("label %q = %q cannot be added, all labels already have values %v", name, value, cmmv.additionalValues)
+			}
+			extended.additionalValues = append(extended.additionalValues, value)
+		}
```
msau42, line 290: I think we should return an error if the value is not specified. Otherwise all your subsequent values will get offset from the label.

metrics/metrics.go
```diff
@@ -101,26 +208,105 @@ var _ CSIMetricsManager = &csiMetricsManager{}
...
+
+// recordMetricsWithLabels is the internal implementation of RecordMetrics.
+func (cmm *csiMetricsManager) recordMetricsWithLabels(
+	operationName string,
+	operationErr error,
+	operationDuration time.Duration,
+	labelValues ...string) {
+	values := []string{cmm.driverName, operationName, getErrorCode(operationErr)}
+	toAdd := len(labelValues)
+	if toAdd > len(cmm.additionalLabelNames) {
```
msau42, line 243: I think we can avoid all of this logic if we just pass in the labels as a map generated in WithLabelValues()

pohly: You are right, that makes the code simpler. Change pushed.

metrics/metrics.go
pohly: I replaced it with a single loop.

metrics/metrics.go
pohly: The intention was to mirror the semantic of https://godoc.org/github.com/prometheus/client_golang/prometheus#SummaryVec.CurryWith which allows adding values repeatedly. I've made that clearer in the function comment.

PR #51, opened 2020-09-07
CHANGELOG-0.8: prepare for v0.8.0

pohly:
Because csi-lib-utils is pre-1.0, we don't need to bump the major
version due to the indirect API change in client-go.

**What type of PR is this?**
/kind documentation

Preparations for v0.8.0.

```release-note
NONE
```

### DIFF
--- a/CHANGELOG-0.8.md
+++ b/CHANGELOG-0.8.md
@@ -1,0 +1,56 @@
+# Release notes for v0.8.0
+
+# Changelog since v0.7.0
+
+## Urgent Upgrade Notes 
+
+### (No, really, you MUST read this before you upgrade)
+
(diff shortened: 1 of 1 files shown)


CHANGELOG-0.8.md
```diff
@@ -0,0 +1,54 @@
+# Changelog since v0.7.0
```
msau42, line 1: Can you also add one more header at the top to make it more obvious what version we're talking about
```
# Release notes for v0.8.0
```

pohly: That's a new thing, right? We didn't do it like that in older changelogs (e.g. https://github.com/kubernetes-csi/external-provisioner/blob/master/CHANGELOG/CHANGELOG-1.6.md).

But I agree, it makes sense. Added.

pohly:
@msau42 can you tag the release after merging?

PR #54, opened 2020-09-29
Add process_start_time_seconds metric into csi metric lib

Jiawei0227:
**What type of PR is this?**
/kind feature

This PR adds the process_start_time_seconds metric into the csi metric lib.

```release-note
Add `process_start_time_seconds` into the csi metric lib.
```

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -181,6 +181,11 @@ func NewCSIMetricsManagerWithOptions(driverName string, options ...MetricsManage
 		subsystem:      SubsystemSidecar,
 		stabilityLevel: metrics.ALPHA,
 	}
+
+	// https://github.com/open-telemetry/opentelemetry-collector/issues/969
+	// Add process_start_time_seconds into the metric to let the start time be parsed correctly
+	metrics.RegisterProcessStartTime(cmm.registry.Register)
+
 	for _, option := range options {
 		option(&cmm)
 	}
@@ -357,7 +362,11 @@ func VerifyMetricsMatch(expectedMetrics, actualMetrics string, metricToIgnore st
 		wantScanner.Scan()
 		wantLine := strings.TrimSpace(wantScanner.Text())
 		gotLine := strings.TrimSpace(gotScanner.Text())
-		if wantLine != gotLine && (metricToIgnore == "" || !strings.HasPrefix(gotLine, metricToIgnore)) {
+		if wantLine != gotLine &&
+			(metricToIgnore == "" || !strings.HasPrefix(gotLine, metricToIgnore)) &&
+			// We should ignore the comments from metricToIgnore, otherwise the verification will
+			// fail because of the comments.
+			!strings.HasPrefix(gotLine, "#") {
 			return fmt.Errorf("\r\nMetric Want: %q\r\nMetric Got:  %q\r\n", wantLine, gotLine)
 		}
 	}
--- a/metrics/metrics_test.go
+++ b/metrics/metrics_test.go
@@ -29,6 +29,11 @@ import (
 	"k8s.io/component-base/metrics/testutil"
 )
 
+const (
+	SidecarOperationMetric = "csi_sidecar_operations_seconds"
+	ProcessStartTimeMetric = "process_start_time_seconds"
+)
+
 func TestRecordMetrics(t *testing.T) {
 	testcases := map[string]struct {
 		subsystem      string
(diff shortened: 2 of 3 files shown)


metrics/metrics.go
```diff
@@ -366,6 +366,7 @@ func VerifyMetricsMatch(expectedMetrics, actualMetrics string, metricToIgnore st
 }
 
 func (cmm *csiMetricsManager) registerMetrics() {
+	metrics.RegisterProcessStartTime(cmm.registry.Register)
```
msau42, line 369: Can you add a comment explaining why the start time metric is needed?

metrics/metrics.go
```diff
@@ -357,7 +357,7 @@ func VerifyMetricsMatch(expectedMetrics, actualMetrics string, metricToIgnore st
 		wantScanner.Scan()
 		wantLine := strings.TrimSpace(wantScanner.Text())
 		gotLine := strings.TrimSpace(gotScanner.Text())
-		if wantLine != gotLine && (metricToIgnore == "" || !strings.HasPrefix(gotLine, metricToIgnore)) {
+		if wantLine != gotLine && (metricToIgnore == "" || !strings.HasPrefix(gotLine, metricToIgnore)) && !strings.HasPrefix(gotLine, "#") {
```
msau42, line 360: Can you add a comment explaining in what scnenario the line won't have a "#"?

metrics/metrics_test.go
```diff
@@ -524,10 +532,10 @@ func TestStartMetricsEndPoint_Noop(t *testing.T) {
 		csi_sidecar_operations_seconds_bucket{driver_name="fake.csi.driver.io",grpc_status_code="OK",method_name="/csi.v1.Controller/ControllerGetCapabilities",le="+Inf"} 1
 		csi_sidecar_operations_seconds_sum{driver_name="fake.csi.driver.io",grpc_status_code="OK",method_name="/csi.v1.Controller/ControllerGetCapabilities"} 20
 		csi_sidecar_operations_seconds_count{driver_name="fake.csi.driver.io",grpc_status_code="OK",method_name="/csi.v1.Controller/ControllerGetCapabilities"} 1
-`
+	`
 
 	actualMetrics := string(contentBytes)
-	if err := VerifyMetricsMatch(expectedMetrics, actualMetrics, ""); err != nil {
+	if err := VerifyMetricsMatch(expectedMetrics, actualMetrics, ProcessStartTimeMetric); err != nil {
```
msau42, line 538: Do we have a unit test that does validate this new metric is output?

Jiawei0227: Added

metrics/metrics.go
Jiawei0227: Done

metrics/metrics_test.go
```diff
@@ -150,8 +157,9 @@ func TestFixedLabels(t *testing.T) {
 		csi_sidecar_operations_seconds_count{a="111",b="222",driver_name="unknown-driver",grpc_status_code="OK",method_name="myOperation"} 1
 	`
 
+	time.Sleep(1000)
```
msau42, line 160: What is the sleep here for?

metrics/metrics.go
pohly: My understanding of https://github.com/kubernetes/component-base/blob/ea4d5eb7ec6256fe0510de29e86ca44f1ac219d2/metrics/processstarttime.go#L37-L48 is that this function should be called once per process.

Calling it here in a low-level utility functions looks wrong to me. It assumes that `registerMetrics()` is called close to the process start. What if the process combines different csiMetricsManager instances under a single endpoint (this is possible, by setting different subsystems)? Won't the two metrics values conflict?

pohly: A better place may be in https://github.com/kubernetes-csi/csi-lib-utils/blob/fb645ec6d10e04a2669c2b6db8c92dac17930133/metrics/metrics.go#L323-L344

pohly (changes requested):

metrics/metrics.go
```diff
@@ -366,6 +370,9 @@ func VerifyMetricsMatch(expectedMetrics, actualMetrics string, metricToIgnore st
 }
 
 func (cmm *csiMetricsManager) registerMetrics() {
+	// https://github.com/open-telemetry/opentelemetry-collector/issues/969
+	// Add process_start_time_seconds into the metric to let the start time be parsed correctly
+	metrics.RegisterProcessStartTime(cmm.registry.Register)
```
pohly, line 375: My understanding of https://github.com/kubernetes/component-base/blob/ea4d5eb7ec6256fe0510de29e86ca44f1ac219d2/metrics/processstarttime.go#L37-L48 is that this function should be called once per process.

Calling it here in a low-level utility function looks wrong to me. It assumes that registerMetrics() is called close to the process start. What if the process combines different csiMetricsManager instances under a single endpoint (this is possible, by setting different subsystems)? Won't the two metrics values conflict?

A better place may be in https://github.com/kubernetes-csi/csi-lib-utils/blob/fb645ec6d10e04a2669c2b6db8c92dac17930133/metrics/metrics.go#L323-L344

msau42: I prefer the constructor: https://github.com/kubernetes-csi/csi-lib-utils/blob/fb645ec6d10e04a2669c2b6db8c92dac17930133/metrics/metrics.go#L178

metrics/metrics_test.go
Jiawei0227: Removed

metrics/metrics.go
Jiawei0227: Hey @pohly , I am not sure I understand the scenario you described. And if they are on the same process then there process_start_time_seconds should be the same? I dont see a problem with it. Maybe I misunderstand.

pohly: At least the fallback (`time.Now().Unix()`) is not constant.

My bigger concern is how registering the same gauge in two different registries will be handled when combining those registries. Will that work?

pohly: > Will that work?

I tried, and apparently it works, so I withdraw that concern.

pohly: > My bigger concern is how registering the same gauge in two different registries will be handled when combining those registries. Will that work?

Sometimes, but not always :disappointed: 

I should have trusted my gut feeling that registering this metric in a low-level utility function looks wrong. After updating dependencies in PMEM-CSI, I now have a conflict between `process_start_time_seconds` registered by csi-lib-utils and some other source for the same metric with a different help text, something that causes Prometheus to fail when collecting metrics with:
```
gathered metric family process_start_time_seconds has help "[ALPHA] Start time of the process since unix epoch in seconds." but should have "Start time of the process since unix epoch in seconds."
```

Apparently that other source is `github.com/prometheus/client_golang/prometheus/process_collector.go`.

It must have worked before when I tested it with two instances that were both registered by csi-lib-utils because the help text was the same.

pohly: Luckily it ended up in the constructor, so making it optional wasn't hard: https://github.com/kubernetes-csi/csi-lib-utils/pull/67

PR #60, opened 2020-10-09
build: update to klog/v2

pohly:
**What type of PR is this?**
/kind cleanup

Although klog and klog/v2 can be used in the same binary, extra work
is needed to ensure that log levels are in sync. It's better to
migrate all components to klog/v2, starting with leaf dependencies
like csi-lib-utils.

```release-note
Projects using csi-lib-utils should update to klog/v2 or must ensure that klog/v1 and klog/v2 are both configured as described in https://github.com/kubernetes/klog/blob/master/examples/coexist_klog_v1_and_v2/coexist_klog_v1_and_v2.go.
```


pohly:
/assign @msau42 

I think we can treat this as a minor update because strictly speaking, no "real" functionality breaks when updating. Log messages might not be visible anymore, depending on the log level settings, when binaries stay with klog/v1.

PR #67, opened 2020-11-11
metrics: optional process_start_time

pohly:
**What type of PR is this?**
/kind bug

The `process_start_time` as registered by csi-lib-utils and thus
Kubernetes component base conflicts with the one from the Prometheus
Golang collector, leading to this error:

```
gathered metric family process_start_time_seconds has help "[ALPHA] Start time of the process since unix epoch in seconds." but should have "Start time of the process since unix epoch in seconds."
```

Allowing processes to control where they get the metric from solves
this conflict.

```release-note
process_start_time should be unique in a process and therefore can now be disabled in the metrics manager registry if not needed or when it conflicts with other collectors
```

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -145,6 +145,22 @@ func WithLabels(labels map[string]string) MetricsManagerOption {
 	}
 }
 
+// WithProcessStartTime controlls whether process_start_time_seconds is registered
+// in the registry of the metrics manager. It's enabled by default out of convenience
+// (no need to do anything special in most sidecars) but should be disabled in more
+// complex scenarios (more than one metrics manager per process, metric already
+// provided elsewhere like via the Prometheus Golang collector).
+//
+// In particular, registering this metric via metric manager and thus the Kubernetes
+// component base conflicts with the Prometheus Golang collector (gathered metric family
+// process_start_time_seconds has help "[ALPHA] Start time of the process since unix epoch in seconds."
+// but should have "Start time of the process since unix epoch in seconds."
+func WithProcessStartTime(registerProcessStartTime bool) MetricsManagerOption {
+	return func(cmm *csiMetricsManager) {
+		cmm.registerProcessStartTime = registerProcessStartTime
+	}
+}
+
 // NewCSIMetricsManagerForSidecar creates and registers metrics for CSI Sidecars and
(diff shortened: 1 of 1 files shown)


Jiawei0227:
/lgtm

There is also another bug fix I opened in k8s for this: https://github.com/kubernetes/kubernetes/pull/96435
Actually I am surprised you can get the process_start_time_seconds in the csi-lib-utils to work. Did you have multiple CSIMetricsManager?

pohly:
> Actually I am surprised you can get the process_start_time_seconds in the csi-lib-utils to work. Did you have multiple CSIMetricsManager?

I think I manually tested with multiple instances before the original PR merged, but that was a one-time thing. I've never actually verified that `process_start_time_seconds` works as intended.

I'm using Prometheus, and that seems to work fine also without it.

Jiawei0227:
> I'm using Prometheus, and that seems to work fine also without it.

Yep, sounds good then. Prometheus will by default register a couple of process related metrics which is good. But all our csi sidecars are only using this registry which dont have this metric. This has been proven problematic because some other OSS library assume we have it and it can cause parsing problem. That's why I added it.

pohly:
> Prometheus will by default register a couple of process related metrics which is good. But all our csi sidecars are only using this registry which dont have this metric. 

I'm wondering whether we should change the sidecars to enable the Prometheus Golang collector.  Does anyone know how useful those metrics are in practice? I don't.

Jiawei0227:
> I'm wondering whether we should change the sidecars to enable the Prometheus Golang collector. Does anyone know how useful those metrics are in practice? I don't.

The thing is we are using our own wrapper around prometheus metrics which is a thing called KubeRegistry in component-base library. And that does not have the prometheus golang collector... here is the issue:
https://github.com/kubernetes/kubernetes/issues/95152

pohly:
> The thing is we are using our own wrapper around prometheus metrics which is a thing called KubeRegistry in component-base library. And that does not have the prometheus golang collector...

That doesn't mean that a sidecar can't add those if it wants. But I understand that it then diverges from the rest of Kubernetes (no definition of metrics stability), so that may not be desirable.

For PMEM-CSI, I enabled it because we have our own definition of "stable features".

PR #68, opened 2020-11-11
Workaround for process start time metric not show up

Jiawei0227:
**What type of PR is this?**
/kind bug

This PR adds a workaround before we upgrade the component-base library dependency. 
The actual fix is here: https://github.com/kubernetes/kubernetes/pull/96435

Fixes #

```release-note
NONE
```

/cc @jsafrane 
/cc @msau42


pohly (changes requested):
Indeed ugly, as you said on chat :sweat_smile: 

But if it gets the job done (have you checked?), then let's use this for now.

Can you add an issue about updating to 1.20 and reverting this?

Jiawei0227:
Yes I checked this time it is correct. The reason why I couldnt capture this when I create the PR is that I got tricked by the go test... This issue is quite tricky so that if you run all the tests together it will not generate any issue. But if you only call one of them it will fail. So that's why I think everything was good before, which is a fault from mine.

Create issue: https://github.com/kubernetes-csi/csi-lib-utils/issues/69

Issue #62, opened 2020-10-19
configurable lease section timers?


danielzhanghl:
hi, 
it seems the CSI lease timers are all hard coded,  could that be enhanced as configurable? 
in case it's used in some slow environment, as other lease? like kube-scheduler/kube-controller-manager

defaultLeaseDuration = 15 * time.Second
defaultRenewDeadline = 10 * time.Second
defaultRetryPeriod   = 5 * time.Second

thanks,

danielzhanghl:
like this PR: https://github.com/kubernetes-csi/csi-lib-utils/pull/75

PR #70, opened 2020-11-12
Add health checker to leader election library

### LINKED ISSUE #66: leader election occasionally fails to reconnect to api server
Exact root cause is still uncertain, but when apiserver is having problems, the csi sidecars will fail to get the leader election lease with this error:

"error retrieving resource lock kube-system/external-attacher-leader-my-driver: Get https://localhost:443/apis/coordination.k8s.io/v1/namespaces/kube-system/leases/external-attacher-leader-my-driver: write tcp [::1]:53540->[::1]:443: write: broken pipe"

Even after apiserver comes back up, this error continues and never recovers. This is apparently intended behavior, and the fix is to enable watchdog so that kubelet can restart the container: https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/healthzadaptor.go#L25

In-tree controllers like kube-controller-manager already set this.

verult:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

This adds leader election health check which reports unhealthy after leadership isn't renewed after some timeout. The CSI controller pod can take advantage of this by setting up a liveness probe for each sidecar container.

Fixes #66

Because the metrics manager already starts a server but only for metrics, this PR refactors the server logic.

The calling sidecar is expected to set up a `http.ServeMux`, pass it to the registration calls in metrics manager and leader election, and listen on the address. [Prototype attacher change](https://github.com/kubernetes-csi/external-attacher/compare/master...verult:le-healthcheck#diff-f422861635f15f707a5f60c81e05a96a541169e64b5f2cae6e0e964f7181b640)

There's a bit of duplicate code across sidecars, but not sure if it's enough to warrant a standalone package.

```release-note
* Added leader election health check.
  * Sidecars should create an HTTP mux (e.g. `http.ServeMux`) and pass it into `RegisterHealthCheck()`.
  * Sidecars are responsible for starting a server with this mux.
  * A liveness probe has to be added to the pod spec for the sidecar container.
* HTTP serving logic for the metrics manager has been refactored.
  * Sidecars should create an HTTP mux (e.g. `http.ServeMux`) and pass it into `RegisterToServer()`.
  * Sidecars are responsible for starting a server with this mux.
```

/assign @msau42 @saad-ali

### DIFF
--- a/leaderelection/leader_election.go
+++ b/leaderelection/leader_election.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"regexp"
 	"strings"
@@ -39,6 +40,14 @@ const (
 	defaultLeaseDuration = 15 * time.Second
 	defaultRenewDeadline = 10 * time.Second
 	defaultRetryPeriod   = 5 * time.Second
+
+	DefaultHealthCheckTimeout = 20 * time.Second
+
+	// HealthCheckerAddress is the address at which the leader election health
+	// checker reports status.
+	// The caller sidecar should document this address in appropriate flag
+	// descriptions.
+	HealthCheckerAddress = "/healthz/leader-election"
 )
 
 // leaderElection is a convenience wrapper around client-go's leader election library.
@@ -55,6 +64,9 @@ type leaderElection struct {
 	// valid options are resourcelock.LeasesResourceLock, resourcelock.EndpointsResourceLock,
 	// and resourcelock.ConfigMapsResourceLock
 	resourceLock string
+	// healthCheck reports unhealthy if leader election fails to renew leadership
+	// within a timeout period.
+	healthCheck *leaderelection.HealthzAdaptor
 
 	leaseDuration time.Duration
 	renewDeadline time.Duration
@@ -134,6 +146,27 @@ func (l *leaderElection) WithContext(ctx context.Context) {
 	l.ctx = ctx
 }
 
+// Server represents any type that could serve HTTP requests for the leader
+// election health check endpoint.
+type Server interface {
+	Handle(pattern string, handler http.Handler)
+}
+
+// PrepareHealthCheck creates a health check for this leader election object
+// with the given healthCheckTimeout and registers its HTTP handler to the given
+// server at the path specified by the constant "healthCheckerAddress".
+// healthCheckTimeout determines the max duration beyond lease expiration
+// allowed before reporting unhealthy.
+// The caller sidecar should document the handler address in appropriate flag
+// descriptions.
+func (l *leaderElection) PrepareHealthCheck(
+	s Server,
+	healthCheckTimeout time.Duration) {
+
+	l.healthCheck = leaderelection.NewLeaderHealthzAdaptor(healthCheckTimeout)
+	s.Handle(HealthCheckerAddress, adaptCheckToHandler(l.healthCheck.Check))
+}
+
 func (l *leaderElection) Run() error {
 	if l.identity == "" {
 		id, err := defaultLeaderElectionIdentity()
@@ -179,6 +212,7 @@ func (l *leaderElection) Run() error {
 				klog.V(3).Infof("new leader detected, current leader: %s", identity)
 			},
 		},
+		WatchDog: l.healthCheck,
 	}
 
 	ctx := l.ctx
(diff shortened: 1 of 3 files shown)


leaderelection/leader_election.go
```diff
@@ -39,6 +40,9 @@ const (
 	defaultLeaseDuration = 15 * time.Second
 	defaultRenewDeadline = 10 * time.Second
 	defaultRetryPeriod   = 5 * time.Second
+	healthCheckTimeout   = 20 * time.Second
+
+	HealthCheckerAddress = "/healthz/leader-election"
```
msau42, line 45: Is there a doc anywhere that shows where this endpoint is used? Is it configurable?

leaderelection/leader_election.go
```diff
@@ -39,6 +40,9 @@ const (
 	defaultLeaseDuration = 15 * time.Second
 	defaultRenewDeadline = 10 * time.Second
 	defaultRetryPeriod   = 5 * time.Second
+	healthCheckTimeout   = 20 * time.Second
```
msau42, line 43: should we make this configurable?

leaderelection/leader_election.go
verult: It's not configurable and I'm leaning towards exposing less knobs until there's a need. The documentation of `RegisterHealthCheck` mentions this constant, although I could explicitly state this URL if you think that could be better. It's up to the caller to document the URL in flag descriptions.

leaderelection/leader_election.go
verult: Leaning towards less knobs until there's a need but no strong preference either way

leaderelection/leader_election.go
msau42: Yeah if you could add a comment just pointing to where it's documented, that would be helpful for someone looking at this.

leaderelection/leader_election.go
msau42: There's at least one ask to make these lease values configurable: https://github.com/kubernetes-csi/csi-lib-utils/issues/62

leaderelection/leader_election.go
verult: Done (if I interpreted your suggestion correctly)

leaderelection/leader_election.go
verult: Done

leaderelection/leader_election.go
msau42: Sorry I meant pointing to where this is defined in the leader election library.

leaderelection/leader_election.go
```diff
@@ -66,16 +78,33 @@ type leaderElection struct {
...
 
 // NewLeaderElection returns the default & preferred leader election type
-func NewLeaderElection(clientset kubernetes.Interface, lockName string, runFunc func(ctx context.Context)) *leaderElection {
-	return NewLeaderElectionWithLeases(clientset, lockName, runFunc)
+// healthCheckTimeout determines the max duration beyond lease expiration
+// allowed before reporting unhealthy.
+func NewLeaderElection(
+	clientset kubernetes.Interface,
+	lockName string,
+	healthCheckTimeout time.Duration,
```
msau42, line 86: Can we use one of the `With*` constructors so we don't need to modify this interface?

danielzhanghl: or read the configuration from a file, if user need to change that, specify that in the file...

verult: Will update to a `With*` constructor. Adding a config file would be out of scope for this PR since a lot of flags could potentially be in a config file instead, though there's also a wider question of, what advantages does a ConfigMap offer over container flags

verult:
Moved timeout param to `PrepareHealthCheck` instead. Also in the latest change, a health check is only created if `PrepareHealthCheck` is called, whereas before all leader election instances would have a health check.

Btw, the endpoint here is `/healthz/leader-election` whereas the one for controller manager is `/healthz/leaderElection`. However, controller manager also has a handler at `/healthz` that aggregates health status from all healthz handlers, and often the liveness probe just points to `/healthz`. I thought having camelCase in an URL is kind of odd (and it might have issues if the binary ever gets run in Windows for some reason), so it's hyphenated here.

leaderelection/leader_election.go
```diff
@@ -71,6 +83,8 @@ func NewLeaderElection(clientset kubernetes.Interface, lockName string, runFunc
 }
 
 // NewLeaderElectionWithLeases returns an implementation of leader election using Leases
+// healthCheckTimeout determines the max duration beyond lease expiration
```
msau42, line 86: remove this comment since it's not specified here. ditto elsewhere

verult: updated

PR #75, opened 2021-01-22
configurable timer for lease

danielzhanghl:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
 /kind feature
> /kind flake

this PR is to make the lease related timer configurable from environment variable.

Fixes #

```release-note
below environment variables could be set: 
         LEASEDURATION:  default 15s
         RENEWDEADLINE: default 10s
         RETRYPERIOD: default 5s

```

### DIFF
--- a/leaderelection/leader_election.go
+++ b/leaderelection/leader_election.go
@@ -43,12 +43,45 @@ const (
 
 	DefaultHealthCheckTimeout = 20 * time.Second
 
+	nameLeaseDuration = "LEASEDURATION"
+	nameRenewDeadline = "RENEWDEADLINE"
+	nameRetryPeriod   = "RETRYPERIOD"
+	nameHealthCheckTimeout = "HEALTHCHECKTIMEOUT"
+
 	// HealthCheckerAddress is the address at which the leader election health
 	// checker reports status.
 	// The caller sidecar should document this address in appropriate flag
 	// descriptions.
 	HealthCheckerAddress = "/healthz/leader-election"
 )
+
+func getTimer(name string) time.Duration {
+
+    envDuration := os.Getenv(name)
+
+    if envDuration != "" {
+        i, err := time.ParseDuration(envDuration)
+        if err == nil {
+            return i
+        }
+    }
+
+    switch name {
+    case nameLeaseDuration:
+        return defaultLeaseDuration
+
+    case nameRenewDeadline:
+        return defaultRenewDeadline
+
+    case nameRetryPeriod:
+        return defaultRetryPeriod
+
+    case nameHealthCheckTimeout:
+        return DefaultHealthCheckTimeout
+    }
+
+    return 0
+}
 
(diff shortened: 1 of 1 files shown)


danielzhanghl:
I signed it

danielzhanghl:
I signed it

msau42:
Hi @danielzhanghl can you describe the problem you're facing with the current lease settings?

/assign @Jiawei0227

danielzhanghl:
the problem is in this ticket, in some environment, the etcd is not that good, leader selection timer need to be configurable to adopt that, as in other K8S component.

https://github.com/kubernetes-csi/csi-lib-utils/issues/62

danielzhanghl:
@Jiawei0227

leaderelection/leader_election.go
```diff
@@ -43,13 +43,46 @@ const (
 
 	DefaultHealthCheckTimeout = 20 * time.Second
 
+	nameLeaseDuration = "LEASEDURATION"
```
Jiawei0227, line 46: Maybe change to 
LEADER_ELECTION_LEASE_DURATION
LEADER_ELECTION_RENEW_DEADLINE
LEADER_ELECTION_RETRY_PERIOD

leaderelection/leader_election.go
```diff
@@ -43,13 +43,46 @@ const (
...
+	nameRenewDeadline = "RENEWDEADLINE"
+	nameRetryPeriod   = "RETRYPERIOD"
+	nameHealthCheckTimeout = "HEALTHCHECKTIMEOUT"
```
Jiawei0227, line 49: Where is this being used?

danielzhanghl: they are used in getTimer(...), except nameHealthCheckTimeout, which is for future and keep the coding style same in this file

leaderelection/leader_election.go
danielzhanghl: name is modeled  from previous const, like "defaultLeaseDuration", would keep it as no change.

leaderelection/leader_election.go
Jiawei0227: Sorry what do you mean for future? When will we use it? Can you share some more details?

leaderelection/leader_election.go
Jiawei0227: These variables are setting from OS environment. So we better be explicitly state what they are used for. Otherwise it can be confused for things like "RETRYPERIOD"

leaderelection/leader_election.go
danielzhanghl: HealthCheckTimeout(DefaultHealthCheckTimeout = 20 * time.Second) is hard coded right now,
if some body think that also need to be configurable,  just change the hard coded value to function call getTimer(nameHealthCheckTimeout), but this is not the issue I met so did not make that change in this package.

PR #77, opened 2021-01-27
Add Migrated metrics to the connection record metrics client side

Jiawei0227:
**What type of PR is this?**
 /kind feature

This PR adds a "migrated" field to the current sidecar operation metrics to indicate if this operation is from a CSI migration PV. For the current sidecar that is using it, it ensure that the new field is not being added. 

To enable this, the sidecar has to create CSIMetricsManager with the correct label or use the `NewCSIMetricsManagerForSidecarWithMigrationStatus()` to create MetricsManager instead.

Partially Addressed # https://github.com/kubernetes/kubernetes/issues/98279 

```release-note
Introduce "migrated" field to metrics for connection library.
```

/cc @pohly @msau42 @saad-ali


connection/connection.go
```diff
@@ -203,11 +210,30 @@ func (cmm ExtendedCSIMetricsManager) RecordMetricsClientInterceptor(
...
-	)
+
+	// record migration status
+	additionalInfo := ctx.Value(AdditionalInfoKey)
+	migrated := "false"
+	if additionalInfo != nil {
+		migrated = additionalInfo.(AdditionalInfo).Migrated
+	}
+	cmmv, metricsErr := cmm.WithLabelValues(map[string]string{metrics.LabelMigrated: migrated})
+	if metricsErr != nil {
```
pohly, line 221: What happens if the sidecar didn't define the "migrated" label? This will cause `WithLabelValues` to fail, right?

I think it would be better to let sidecars opt into this feature. This means only calling `WithLabelValues` if there is some non-empty value to be recorded.

connection/connection.go
```diff
@@ -203,11 +210,30 @@ func (cmm ExtendedCSIMetricsManager) RecordMetricsClientInterceptor(
...
+		klog.V(5).Infof("Failed to record migrated status, error: %v", metricsErr)
```
pohly, line 222: IMHO this should cause the wrapper to fail. A log message with `V(5)` is likely to go unnoticed and this shouldn't happen in practice.

metrics/metrics.go
```diff
@@ -175,6 +178,17 @@ func NewCSIMetricsManagerForSidecar(driverName string) CSIMetricsManager {
...
 }
 
+// NewCSIMetricsManagerForSidecarWithMigrationStatus creates and registers metrics
+// for CSI Sidecars and returns an object that can be used to trigger the metrics.
+// It uses "csi_sidecar" as subsystem. Meanwhile, it adds the migrated label to
+// the metrics to indicate if the operation is from a csi migration
+//
+// driverName - Name of the CSI driver against which this operation was executed.
+//              If unknown, leave empty, and use SetDriverName method to update later.
+func NewCSIMetricsManagerForSidecarWithMigrationStatus(driverName string) CSIMetricsManager {
```
pohly, line 188: I don't think we need this helper function. Just document how to enable this feature and then let sidecars call `NewCSIMetricsManagerWithOptions(driverName, WithLabelNames(LabelMigrated))`.

pohly (changes requested):

connection/connection.go
Jiawei0227: If the sidecar did not define the migrated label, we will fall back to use the old cmm to record metrics. So the old sidecar metrics will still be the same. See comments on line 229

metrics/metrics.go
Jiawei0227: Sure thing. I will remove it.

connection/connection.go
Jiawei0227: Sorry I was adding comments to the wrong section. Now this should make sense. We have two cases here:
1. The metricsManager does not define labelMigration, we will fail the `WithLabelValues` with metricsErr and so we fall back to use the original metric. I dont know in this condition if we should log something... This can happen when a sidecar upgrade to the newest csi-lib-util version and did not change there MetricsManager initialization. What's your thought? 
2. The metricsManager has the labelMigration, the metricsErr will be nil, we use the wrapper to record the metric.

metrics/metrics.go
```diff
@@ -47,6 +47,9 @@ const (
 	labelGrpcStatusCode   = "grpc_status_code"
 	unknownCSIDriverName  = "unknown-driver"
 
+	// LabelMigrated is the Label that indicate whether this is a migrated operation
```
msau42, line 50: maybe call it "CSI migration" so it's more obvious what "migrated" means

connection/connection.go
msau42: I'm a little confused by this fallback behavior. Shouldn't we define "migrated" in the metric schema?

Jiawei0227: > I'm a little confused by this fallback behavior. Shouldn't we define "migrated" in the metric schema?

That would also work... But that means we will have the migrated field in the MetricsManager by default and everything that is using it will have this new field, which I do not know if it is something we want... This current way of using `NewCSIMetricsManagerWithOptions(driverName, WithLabelNames(LabelMigrated))` is more like an opt-in way. So a sidecar/plugin can decide if they want this field or not...

metrics/metrics.go
Jiawei0227: Done

connection/connection.go
Jiawei0227: I change the HaveAdditionalLabel to public so hopefully the logic makes more sense here. 
So basically we will check if the metricsManager has the migrated label. If it does, it will record the migrated status, otherwise it will just use the old metrics.

msau42:
Is NewCSIMetricsManagerForSidecarWithMigrationStatus defined?

Jiawei0227:
> Is NewCSIMetricsManagerForSidecarWithMigrationStatus defined?

I originally had it, and it seems @pohly think it is unnecessary. I think we can just add it here and it does not hurt if people are not using it.

Jiawei0227:
> Is NewCSIMetricsManagerForSidecarWithMigrationStatus defined?

Done

pohly:
> >    Is NewCSIMetricsManagerForSidecarWithMigrationStatus defined?
>
> I originally had it, and it seems @pohly think it is unnecessary. I think we can just add it here and it does not hurt if people are not using it.

My concern is that this will get out of hand. What's next, `NewCSIMetricsManagerForSidecarWithMigrationStatusAndAlsoThisOtherThing`? 

`NewCSIMetricsManagerForSidecar` makes sense to me as "give me the metrics manager for sidecars with the default options".

How about we define what those default options are and then a sidecar that wants to the default options plus something else can use that slice plus the additional ones when calling `NewCSIMetricsManagerWithOptions`?

msau42:
Could a builder work here? Something like 'SidecarMetrics().WithMigration()'

Jiawei0227:
> Could a builder work here? Something like 'SidecarMetrics().WithMigration()'

I create a `WithMigration()` metricsOption. We originally did not choose to use the builder model. And instead go with the option way. So I think an option looks more reasonable here.

pohly:
> How about we define what those default options are and then a sidecar that wants to the default options plus something else can use that slice plus the additional ones when calling NewCSIMetricsManagerWithOptions?

Here's a quick draft what that would look like:
```
func NewCSIMetricsManagerForSidecar(driverName string) CSIMetricsManager {
	return NewCSIMetricsManagerWithOptions(driverName, DefaultSidecarOptions()...)
}

// DefaultSidecarOptions returns the default options for a metrics manager that is used
// by a sidecar. Currently it is empty, but sidecars that call NewCSIMetricsManagerWithOptions
// should start with these options and then add their own, custom ones in case that
// some default options get added in a future release.
func DefaultSidecarOptions() []MetricsManagerOption {
   return nil
}

func NewCSIMetricsManagerForPlugin(driverName string) CSIMetricsManager {
	return NewCSIMetricsManagerWithOptions(driverName, DefaultPluginOptions()...)
}

func DefaultPluginOptions() []MetricsManagerOption {
    return []MetricsManagerOption{
       WithSubsystem(SubsystemPlugin),
   }
}
```

And then in a sidecar which would have used `NewCSIMetricsManagerForSidecarWithMigrationStatus`:
```
  cmm := NewCSIMetricsManagerWithOptions(driverName, append(DefaultSidecarOptions(), WithMigration())...)
```

Jiawei0227:
> And then in a sidecar which would have used NewCSIMetricsManagerForSidecarWithMigrationStatus:

I am wondering if we should put migration status into the default option for sidecar.

pohly:
> I am wondering if we should put migration status into the default option for sidecar.

I'm not sure. If a sidecar never modifies the context, then it also won't need the metric pre-defined.

Jiawei0227:
> I'm not sure. If a sidecar never modifies the context, then it also won't need the metric pre-defined.

If it does not modify context it will get "false" as the value. The reason why I think this can be the default is that we might not want to have different schema for sidecars. It will be hard for SRE to consume. So for the sidecars that does not have migration like snapshotter, we just default the value to false. What do you think?

Jiawei0227:
> I'm not sure. If a sidecar never modifies the context, then it also won't need the metric pre-defined.

Or I think we can leave the `WithMigration()` implementation as it is. And do not add the default option thing until we feel there is a need? Otherwise I dont see the value we are adding by having the DefaultSidecarOptions()

msau42:
> If it does not modify context it will get "false" as the value. The reason why I think this can be the default is that we might not want to have different schema for sidecars. It will be hard for SRE to consume. So for the sidecars that does not have migration like snapshotter, we just default the value to false. What do you think?

In terms of backwards compatibility, it is fine for a metric to have less fields than the metric schema defined in the collector. So if some sidecars don't have all the fields, or if some plugins don't support migration, it should be ok. What will fail is if the actual metric has more fields than what is defined in the collector.

Jiawei0227:
> In terms of backwards compatibility, it is fine for a metric to have less fields than the metric schema defined in the collector. So if some sidecars don't have all the fields, or if some plugins don't support migration, it should be ok. What will fail is if the actual metric has more fields than what is defined in the collector.

In that case, I think the current PR should be good. Unless there are any other default fields we need to add to the sidecar, there is no need to add the DefaultSidecarOptions() for now.

connection/connection.go
```diff
@@ -203,11 +210,33 @@ func (cmm ExtendedCSIMetricsManager) RecordMetricsClientInterceptor(
...
+		additionalInfo := ctx.Value(AdditionalInfoKey)
+		migrated := "false"
+		if additionalInfo != nil {
+			migrated = additionalInfo.(AdditionalInfo).Migrated
+		}
+		cmmv, metricsErr := cmm.WithLabelValues(map[string]string{metrics.LabelMigrated: migrated})
+		if metricsErr != nil {
+			klog.Errorf("Failed to record migrated status, error: %v", metricsErr)
+		} else {
+			cmmv.RecordMetrics(
```
msau42, line 225: Can we save a few lines and have both paths share this call?

msau42 (commented):
/approve

connection/connection.go
Jiawei0227: I would love to but the caller is different... The cmmv is a different thing with cmm. Do you have any suggestion on how can I combine these two?

msau42: Instead of creating a new cmmv, can you set the return value to cmm?

Jiawei0227: Done

PR #95, opened 2021-07-27
Move CSI secret handling logic to library

### LINKED ISSUE #94: Move CSI secret handling logic to library
As CSI secret handling logic can be [used](https://github.com/kubernetes-csi/external-provisioner/issues/653) by multiple controllers, it should be moved to this repository.

mkimuram:
**What type of PR is this?**
/kind cleanup

This PR moves parameters/secret related functions from "kubernetes-csi/external-provisioner" and 'kubernetes-csi/external-snapshotter" to "csi-lib-utils". It is needed because these functions are expected to be used by other controllers.

Fixes #94

```release-note
A new "params" package is added and below functions are provided:

- VerifyAndGetSecretNameAndNamespaceTemplate()
- GetSnapshotSecretReference()
- RemovePrefixedSnapshotParameters()
- GetVolumeSecretReference()
- RemovePrefixedParameters()

This package also provides "SecretParamsMap" struct that has GetSecretNameKey() and GetSecretNamespaceKey() methods. As "SecretParamsMap", below variables are provided:

For volume:
- ProvisionerSecretParams
- NodePublishSecretParams
- ControllerPublishSecretParams 
- NodeStageSecretParams
- ControllerExpandSecretParams
For snapshot:
- SnapshotterSecretParams
- SnapshotterListSecretParams
```


mkimuram:
@msau42 @xing-yang 

Still not tested "kubernetes-csi/external-provisioner" and 'kubernetes-csi/external-snapshotter" with this version of library. Sharing this to get early feedback on the package structure and the design of functions/parameters.

The points that need feedback are:
- Remove*Parameters functions are also moved to this package because *Params `SecretParamsMap` are tightly bound to this method. Is this design OK? (Without these functions, package name was expected to be "secret". Alternatively, we might expose all Const values as public scoped).
- By moving volume related functions and snapshot related functions to the same package, dependency to "github.com/kubernetes-csi/external-snapshotter/client/v4" package is introduced even if a user doesn't need to use snapshot. Is it OK or is it better to move to separate package to avoid it?
- `VerifyAndGetSecretNameAndNamespaceTemplate()` and `SecretParamsMap`.{`GetSecretNameKey()`, `GetSecretNamespaceKey()`} are exposed as public scoped, which are expected to be used in https://github.com/kubernetes-csi/external-provisioner/issues/654. Is it OK or should it be in a separate PR?

xing-yang:
I'm concerned that this lib is getting too heavy by adding dependency on snapshot client/v4.

mkimuram:
@xing-yang

Thank you for sharing your concern.
As you mentioned, it seems that dependency is added on using the library even if the package, or 'params' package in this case, isn't used. As seen below:

```
mkdir -p deptest
cd deptest/
cat << EOF > deptest.go 
package main

import "github.com/kubernetes-csi/csi-lib-utils/deprecatedflags"

var _ = deprecatedflags.Add("foo")

func main() {
}
EOF

go mod init
go mod tidy
echo "replace github.com/kubernetes-csi/csi-lib-utils => ../csi-lib-utils" >> go.mod
go mod tidy

grep snapshot go.sum
github.com/kubernetes-csi/external-snapshotter/client/v4 v4.1.0/go.mod h1:YBCo4DoEeDndqvAn6eeu0vWM7QdXmHEeI9cFWplmBys=
```

I will consider a way to avoid passing `VolumeSnapshot` struct.
It just seems to need below information, not the entire struct. (Just passing `snapshot.Name` and `snapshot.Namespace` will be fine?)
```
namespaceParams := map[string]string{
	"volumesnapshotcontent.name": snapContentName,
	"volumesnapshot.namespace": snapshot.Namespace,
}
map[string]string{
	"volumesnapshotcontent.name": snapContentName,
	"volumesnapshot.name": snapshot.Name,
	"volumesnapshot.namespace": snapshot.Namespace,
}
```

mkimuram:
Above commit removes the dependency. The same approach can be applied to Volume. As for volume, it needs to pass `pvc.namespace`, `pvc.name`, and `pvc.annotation`. Should it also be changed?

WDYT?

pohly:
> Above commit removes the dependency. The same approach can be applied to Volume. As for volume, it needs to pass `pvc.namespace`, `pvc.name`, and `pvc.annotation`. Should it also be changed?

Depending on `v1.PersistentVolumeClaim` shouldn't be a problem, but perhaps change it for the sake of consistency?

mkimuram:
> Depending on v1.PersistentVolumeClaim shouldn't be a problem, but perhaps change it for the sake of consistency?

Yes, just for consistency. It's not a big difference, but it's a public scoped interface, which will be difficult to change later. So, I just would like to be careful in design decision.

pohly:
I prefer pvc.namespace, pvc.name, and pvc.annotation as separate parameters, partly for consistency, partly because clearly documents which fields are needed. The latter is less clear when passing a PVC.

PR #109, opened 2022-03-03
leader election: remove obsolete ConfigMapsResourceLock and EndpointsResourceLock

pohly:
**What type of PR is this?**

/kind cleanup

Kubernetes 1.24 won't support those anymore. To support building with a newer
client-go, csi-lib-utils must not reference the removed constants. This doesn't
affect the sidecars because those only use the default NewLeaderElection.

See https://github.com/kubernetes/kubernetes/commit/29d9683cd0fb3cc81810a8b39715e5eb4c68b00e

```release-note
Leader election is only supported for leases via NewLeaderElection.
```

### DIFF
--- a/leaderelection/leader_election.go
+++ b/leaderelection/leader_election.go
@@ -61,8 +61,7 @@ type leaderElection struct {
 	// the namespace to store the lock resource
 	namespace string
 	// resourceLock defines the type of leaderelection that should be used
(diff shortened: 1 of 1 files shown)


ggriffiths:
Change makes sense, looks good.

/lgtm

ggriffiths (approved):

PR #112, opened 2022-05-02
Automated cherry pick of #81: connection: exit with less output on connection loss

ialidzhikov:
/sig storage
/kind bug

Cherry pick of #81 on release-0.9.

#81: connection: exit with less output on connection loss

Part of https://github.com/kubernetes-csi/external-provisioner/issues/732

```release-note
sidecars exit without output about goroutines when losing the connection to the driver
```

For details on the cherry pick process, see the [cherry pick requests](https://git.k8s.io/community/contributors/devel/sig-release/cherry-picks.md) page.

### DIFF
--- a/connection/connection.go
+++ b/connection/connection.go
@@ -84,7 +84,8 @@ func ExitOnConnectionLoss() func() bool {
 		if err := ioutil.WriteFile(terminationLogPath, []byte(terminationMsg), 0644); err != nil {
 			klog.Errorf("%s: %s", terminationLogPath, err)
 		}
-		klog.Fatalf(terminationMsg)
+		klog.Exit(terminationMsg)
+		// Not reached.
 		return false
 	}
 }


ialidzhikov:
> Can we update the older external-provisioner branches to the latest csi-lib-utils instead?

Sure, this is also an option. I was not sure how backwards-compatible is updating minor versions of this library from 1 patch version to another one for a sidecar (for example external-provisioner). That's why the safer approach to me is to have v0.9.2 with this concrete patch. WDYT?

pohly:
I looked at the changes in csi-lib-utils 0.10 and 0.11 and I think those are safe all the way back to release-2.1.

ialidzhikov:
Okay, then I will close this PR. Thanks!

/close

ialidzhikov:
@pohly I quickly tried out. In short github.com/kubernetes-csi/csi-lib-utils@v0.10.0 has its K8s dependencies updated to v0.22.0. Trying to vendor github.com/kubernetes-csi/csi-lib-utils@v0.10.0 in external-provisioner@release-2.2 does a lot of changes under vendor:

```
$ git diff --shortstat
 193 files changed, 1923 insertions(+), 4620 deletions(-)
```

(dependencies like `golang.org/x/net`, `golang.org/x/sys`, `google.golang.org/grpc` get updated)

I would suggest to pick the approach with the cherry-pick an new patch release for github.com/kubernetes-csi/csi-lib-utils@v0.9.2. Looks cleaner and safer to me.

pohly (approved):
/lgtm
/approve

PR #113, opened 2022-05-06
update client-go and other dependencies to latest version

humblec:
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>

> /kind cleanup

-->
```release-note
NONE
```


pohly:
I have my doubts whether we need to bump dependencies in csi-lib-utils. We are not building anything from it, and projects that consume it can update their dependencies themselves.

Because we bumped the client-go dependencies, we almost had to fork csi-lib-utils for older release branches because that would have forced those to also update the deps.

There is no particular reason for this PR, right?

PR #118, opened 2022-09-01
add WithCustomRegistry for metrics

ayanamist:
**What type of PR is this?**

> /kind feature

If users want to monitor workqueue metrics, they usually import `k8s.io/component-base/metrics/prometheus/workqueue`, which registers to `k8s.io/component-base/metrics/legacyregistry`, however, `csiMetricsManager` is using its own newly-created registry.

Make a new method `WithCustomRegistry` to let users pass in an exist registry.

```release-note
Adds WithCustomRegistry method to the metrics pkg to allow reuse of an existing registry.
```

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -179,6 +179,13 @@ func WithProcessStartTime(registerProcessStartTime bool) MetricsManagerOption {
 	}
 }
 
+// WithCustomRegistry allow user to use custom pre-created registry instead of a new created one.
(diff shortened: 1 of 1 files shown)


ggriffiths:
I don't see any issues with this change. Can you clarify the release note though? Something like this-

```
Adds WithCustomRegistry method to the metrics pkg to allow reuse of an existing registry.
```

ggriffiths (approved):
/lgtm

sunnylovestiramisu:
Is this compatible with k8s.io/component-base v0.26.0?

PR #123, opened 2022-12-19
Update dependency go modules for k8s v1.26.0

sunnylovestiramisu:
Ran kubernetes-csi/csi-release-tools go-get-kubernetes.sh -p 1.26.0


sunnylovestiramisu:
``make test`` failed with TestConnectMetrics

The diffs are:
```
-csi_sidecar_operations_seconds_sum{driver_name="fake.csi.driver.io",grpc_status_code="Unimplemented",method_name="/csi.v1.Identity/GetPluginInfo"} 0
        +csi_sidecar_operations_seconds_sum{driver_name="fake.csi.driver.io",grpc_status_code="Unimplemented",method_name="/csi.v1.Identity/GetPluginInfo"} 0.000649487

=======================

-csi_plugin_operations_seconds_sum{driver_name="fake.csi.driver.io",grpc_status_code="Unimplemented",method_name="/csi.v1.Identity/GetPluginInfo"} 0
        +csi_plugin_operations_seconds_sum{driver_name="fake.csi.driver.io",grpc_status_code="Unimplemented",method_name="/csi.v1.Identity/GetPluginInfo"} 1.075e-06

=======================

-csi_sidecar_operations_seconds_sum{driver_name="fake.csi.driver.io",grpc_status_code="Unimplemented",method_name="/csi.v1.Identity/GetPluginInfo",migrated="true"} 0
        +csi_sidecar_operations_seconds_sum{driver_name="fake.csi.driver.io",grpc_status_code="Unimplemented",method_name="/csi.v1.Identity/GetPluginInfo",migrated="true"} 0.000252969

```

Seems like the type conversion in between changed from int to some doubles?

sunnylovestiramisu:
Also noticed that the unit test is saying we should ignore the csi_sidecar_operations_seconds_sum metrics?

https://github.com/kubernetes-csi/csi-lib-utils/blob/master/connection/connection_test.go#L429

sunnylovestiramisu:
/assign @Jiawei0227 

Do you have any context on this?

sunnylovestiramisu:
1.25.0 release is not broken, the breaking change started at 1.26.0:
https://github.com/kubernetes-csi/csi-lib-utils/pull/124

sunnylovestiramisu:
I am not sure what the magic is but i copied the dependencies from the 1.25.0 pr and then ran go mod tidy/vendor etc. and now this pr is passing as well.

go.mod
```diff
@@ -1,23 +1,62 @@
...
 require (
 	github.com/container-storage-interface/spec v1.5.0
 	github.com/golang/protobuf v1.5.2
-	github.com/stretchr/testify v1.7.0
-	golang.org/x/net v0.0.0-20210520170846-37e1c6afe023
-	google.golang.org/grpc v1.38.0
-	k8s.io/api v0.22.0
-	k8s.io/client-go v0.22.0
+	github.com/stretchr/testify v1.8.0
+	golang.org/x/net v0.3.1-0.20221206200815-1e63c2f08a10
```
xing-yang, line 9: Can you change this to v0.4.0?  There's a CVE in earlier version.

go.mod
```diff
@@ -1,23 +1,62 @@
...
+	google.golang.org/grpc v1.49.0
+	k8s.io/api v0.26.0
+	k8s.io/client-go v0.26.0
 	k8s.io/component-base v0.22.0
-	k8s.io/klog/v2 v2.9.0
+	k8s.io/klog/v2 v2.80.1
```
xing-yang, line 14: k8s.io/component-base v0.22.0?

Can this be updated to 0.26.0?

go.mod
sunnylovestiramisu: Sure let me do that

go.mod
```diff
@@ -1,23 +1,62 @@
 module github.com/kubernetes-csi/csi-lib-utils
 
-go 1.16
+go 1.19
 
 require (
 	github.com/container-storage-interface/spec v1.5.0
```
xing-yang, line 6: Can you update this to 1.7.0?

go.mod
sunnylovestiramisu: Update to 0.26.0 break the unit tests as we see above.

go.mod
sunnylovestiramisu: Why 1.17? k8s 1.26.0 is at 1.19

xing-yang: github.com/container-storage-interface/spec v1.5.0 -> v1.7.0 which is the latest CSI spec release.  This is not K8s release.

msau42: this comment is about github.com/container-storage-interface/spec

go.mod
xing-yang: I see.  Can you add a comment about this in the PR description?

msau42: I think we should update the component-base and fix the tests. I'm not sure if there will be any side effects if the sidecars and the library use two different versions of component-base: https://github.com/kubernetes-csi/external-provisioner/blob/master/go.mod#L24

go.mod
sunnylovestiramisu: Then I need to fix the previous 1.25.0 first

sunnylovestiramisu: But the go version for github.com/container-storage-interface is 1.18? https://github.com/container-storage-interface/spec/blob/release-1.7/go.mod

xing-yang: This should be fine.  It's not possible for all dependency libs to use the exact same go version.

go.mod
```diff
@@ -1,34 +1,33 @@
 module github.com/kubernetes-csi/csi-lib-utils
 
-go 1.18
+go 1.17
```
xing-yang, line 3: This should still go to 1.19.

go.mod
xing-yang: I see this is still `github.com/container-storage-interface/spec v1.5.0`.  Is there any issue updating to 1.7.0?

go.mod
xing-yang: You are going to update this to v0.4.0, right?

go.mod
sunnylovestiramisu: Oh sorry I somehow was thinking you want me to update the go version to 1.17.0, changing it now. And i am still trying to fix the component base.

go.mod
```diff
@@ -1,34 +1,33 @@
 module github.com/kubernetes-csi/csi-lib-utils
 
-go 1.18
+go 1.19
```
pohly, line 3: Please keep it at 1.18. 1.19 would indicate that the code in this repo needs 1.19, which isn't the case.

sunnylovestiramisu:
The error has changed from this [line](https://github.com/kubernetes-csi/csi-lib-utils/blob/master/connection/connection_test.go#L450) after upgraded to component-base 0.26.0

```
Diff:
--- metric output does not match expectation; want
+++ got:
@@ -16,3 +16,3 @@
 csi_plugin_operations_seconds_bucket{driver_name="fake.csi.driver.io",grpc_status_code="Unimplemented",method_name="/csi.v1.Identity/GetPluginInfo",le="+Inf"} 1
-csi_plugin_operations_seconds_sum{driver_name="fake.csi.driver.io",grpc_status_code="Unimplemented",method_name="/csi.v1.Identity/GetPluginInfo"} 0
+csi_plugin_operations_seconds_sum{driver_name="fake.csi.driver.io",grpc_status_code="Unimplemented",method_name="/csi.v1.Identity/GetPluginInfo"} 4.81e-07
 csi_plugin_operations_seconds_count{driver_name="fake.csi.driver.io",grpc_status_code="Unimplemented",method_name="/csi.v1.Identity/GetPluginInfo"} 1
connection_test.go:458: 
```

This results in the want split does not have the expected string to compare on this [line](https://github.com/kubernetes-csi/csi-lib-utils/blob/master/connection/connection_test.go#L464)

The want string is:
```
Diff:
--- metric output does not match expectation; want
+++ 
```

We need to change the split rule to something without the ":":
```
wantArr := strings.Split(err.Error(), "want")
```

This also ends up in certain situations the want string contains one line extra, so we need to remove that line so we can scan later:

```
want = strings.ReplaceAll(want, "+++ got:", "")
```

go.mod
sunnylovestiramisu: This one has been constantly downgraded again. Let me update.

PR #127, opened 2023-01-13
add option to limit grpc logging length

leiyiz:
**What type of PR is this?**
/kind feature

in some cases the GRPC response is huge and logging it entirely causes log space to fill up immediately.

Fixes #

```release-note
Now there's an option to set a universal limit on GRPC log length.
```

### DIFF
--- a/connection/connection.go
+++ b/connection/connection.go
@@ -19,6 +19,7 @@ package connection
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"net"
 	"strings"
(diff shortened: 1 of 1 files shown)


leiyiz:
this is similar to how pdcsi driver is going to limit log length in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1085

msau42 (commented):
Log length is mainly an issue with List volume and snapshot calls, where there could be a lot of objects in the system.

The main downside I can see with this approach is that if there are any important fields in the struct after the list of volumes/snapshots, then those will get dropped. But in practice, that is not relevant to the csi list calls.

Another approach could be to allow setting a custom log function, and then having attacher and snapshotter implement some custom handling where maybe it could treat List calls at a debug log level. WDYT? Do you see value in logging the first few entries vs not logging any entries at all?

leiyiz:
I think with truncated logging we can potentially do some driver-side randomization on the sequence of the list of Volumes in the future thus retain some debuggability. With the custom logger approach my intuition is that there's more potential to omit useful information.

msau42:
Thanks, I think this is sufficient for now. If anyone else thinks a custom logger may be beneficial, we can look into that as another enhancement.

/lgtm
/approve

PR #128, opened 2023-01-18
Automated cherry pick of #127: add option to limit grpc logging length

leiyiz:
Cherry pick of #127 on release-0.12.

#127: add option to limit grpc logging length

For details on the cherry pick process, see the [cherry pick requests](https://git.k8s.io/community/contributors/devel/sig-release/cherry-picks.md) page.

```release-note

```

### DIFF
--- a/connection/connection.go
+++ b/connection/connection.go
@@ -19,6 +19,7 @@ package connection
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"net"
 	"strings"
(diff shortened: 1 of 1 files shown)


jsafrane:
Does it make sense to backport this change? It does not look like a bugfix to me.

leiyiz:
I guess what I'm trying to do is to release a new version of `csi-lib-utils` to include this and it seemed fit to do a `0.12.1` version because it is not introducing any breaking changes.

msau42:
I think it's ok to release a 0.13.0 and use that in csi-attacher

PR #131, opened 2023-05-10
Add timeout to connection.Connect()

ConnorJC3:
**What type of PR is this?**

/kind bug

If connection.Connect() gets stuck it will block infinitely, never reporting an error to the library's consumer.

We have received reports of this happening even when a _valid_ address is passed in node-driver-registrar because of a race condition involving pod restarts. Specifically, there is an unlucky sequence of events when restarting both the CSI Driver and node-driver-registrar, wheere the node-driver-registar will attempt to Connect() to the CSI Driver, but will get stuck doing so on an old file descriptor from the previously running CSI Driver (and thus, get stuck infinitely).

There is no mechanism to pass a connection timeout to Connect, so this commit adds a reasonbly long default timeout so these cases will eventually return an error rather than getting stuck infinitely.

```release-note
NONE
```

### DIFF
--- a/connection/connection.go
+++ b/connection/connection.go
@@ -54,8 +54,8 @@ func SetMaxGRPCLogLength(characterCount int) {
 // file or have format '<protocol>://', following gRPC name resolution mechanism at
 // https://github.com/grpc/grpc/blob/master/doc/naming.md.
 //
-// The function tries to connect indefinitely every second until it connects. The function automatically disables TLS
-// and adds interceptor for logging of all gRPC messages at level 5.
+// The function tries to connect for 30 seconds, and returns an error if no connection has been established at that point.
+// The function automatically disables TLS and adds interceptor for logging of all gRPC messages at level 5.
 //
 // For a connection to a Unix Domain socket, the behavior after
 // loosing the connection is configurable. The default is to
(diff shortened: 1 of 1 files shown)


connection/connection.go
```diff
@@ -70,7 +70,7 @@ func SetMaxGRPCLogLength(characterCount int) {
 // For other connections, the default behavior from gRPC is used and
 // loss of connection is not detected reliably.
 func Connect(address string, metricsManager metrics.CSIMetricsManager, options ...Option) (*grpc.ClientConn, error) {
-	return connect(address, metricsManager, []grpc.DialOption{}, options)
+	return connect(address, metricsManager, []grpc.DialOption{grpc.WithTimeout(time.Second * 30)}, options)
```
msau42, line 73: The comment should also be updated.

xing-yang:
@ConnorJC3 Can you address the comments?  We're going to cut a release soon.  Thanks.

ConnorJC3:
@xing-yang sorry, I somehow missed the above review comment. New version of the PR uploaded with the comment changed.

connection/connection.go
```diff
@@ -54,8 +54,8 @@ func SetMaxGRPCLogLength(characterCount int) {
 // file or have format '<protocol>://', following gRPC name resolution mechanism at
 // https://github.com/grpc/grpc/blob/master/doc/naming.md.
 //
-// The function tries to connect indefinitely every second until it connects. The function automatically disables TLS
-// and adds interceptor for logging of all gRPC messages at level 5.
+// The function tries to connect for 30 seconds, and returns an error if no connection has been establashed at that point.
```
xing-yang, line 57: typo: establashed

ConnorJC3: Fixed

PR #135, opened 2023-06-25
chore: use String() instead of fmt.Sprintf

testwill:
**What type of PR is this?**
> /kind cleanup

should use String() instead of fmt.Sprintf (S1025)

Fixes #

```release-note
NONE
```


pohly (commented):
/ok-to-test

PR #140, opened 2023-07-31
Add support for otel tracing of grpc calls

### LINKED ISSUE #139: Add otel trace instrumentation on gRPC calls
I'm proposing to add some basic tracing instrumentation of the gRPC calls made by the CSI clients using the code in [connection.go](https://github.com/kubernetes-csi/csi-lib-utils/blob/master/connection/connection.go) to connect to the CSI driver. 
The feature simply consists in adding the [otelgrpc.UnaryClientInterceptor()](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc#UnaryClientInterceptor) to the existing [ChainUnaryInterceptor](https://github.com/kubernetes-csi/csi-lib-utils/blob/master/connection/connection.go#L121). This is enough to create traces for all gRPC calls and they can be easily exported by the client.

I think it can be done by creating a new `ConnectWithGrpcInterceptor` function that will call the `connect` function with the added interceptor as a `DialOption`. This way the feature will be opt-in for the users.

I can contribute it (and the implementation in the CSI components) if it is something the community wants to see. I already implemented something similar on the [azuredisk-csi-driver](https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1912) and plan to do the same for aws and gcp drivers.

Fricounet:
**What type of PR is this?**

/kind feature

- Refactor `Connect` to rely more on Options to configure how the gRPC client connect to the server. Configuring how `connect` behaves relies on new functions like `withMetrics` or `withTimeout` that need to be passed as options at the caller level. This will make it easier to add new configuration options without breaking the api in the future.
- Create a new `WithOtelTracing` option to be used by gRPC client when connecting to the CSI server with `Connect`. This function adds a gRPC interceptor that will provide basic tracing of the gRPC calls.
Once it is merged and released, I will go through the different CSI sidecars to allow the use of this option behind a feature flag `--enable-otel-tracing`.

Below is an example of what it looks like when using this change on a CSI sidecar. I have otel tracing activated on the CSI driver (server) and the sidecar (client). We can see a trace composed of 2 spans with their duration.
image

Fixes #139 

I split the PR in 2 commits (dependency change first and code change next) to make it easier to review. If you prefer to only merge one commit, I can squash commits once the review is done.

```release-note
Added WithOtelTracing option to connect to the CSI driver gRPC server with otel gRPC tracing interceptor
```


Fricounet:
Hey @gnufied, I see you did a PR recently #143 to allow disabling metrics on gRPC connection. You did something fairly similar to what I wanted to do here to add otel tracing. This made me realise that this approach lacks flexibility because we won't be able to have a gRPC connection with tracing without metrics for instance. Unless we write a separate function for each case which does not seem like a good idea.

So I think this part need to be refactored. I was thinking about using the `Option` type already available and create functions like `withMetrics` and `WithOtelGrpcInterceptor` that could be added as parameters when calling `connect`.
Without breaking the existing API, it could look something like this:
```go 
func Connect(address string, metricsManager metrics.CSIMetricsManager, options ...Option) (*grpc.ClientConn, error) {
        if metricsManager != nil {
                options = append(options, withMetrics(metricsManager))
        }
	return connect(address, []grpc.DialOption{grpc.WithTimeout(time.Second * 30)}, options)
}
```

For your use case, instead of doing: `ConnectWithoutMetrics(address)`, you would do `Connect(address, nil)`. What do you think about it?

gnufied:
Sure, Feel free to refactor your PR to design you proposed. And lets make sure this handles tracing changes you are introducing here too.

connection/connection.go
```diff
@@ -70,12 +72,11 @@ func SetMaxGRPCLogLength(characterCount int) {
...
-
-// ConnectWithoutMetrics behaves exactly like Connect except no metrics are recorded.
-func ConnectWithoutMetrics(address string, options ...Option) (*grpc.ClientConn, error) {
-	return connect(address, nil, []grpc.DialOption{grpc.WithTimeout(time.Second * 30)}, options)
+	options = append(options, withTimeout(time.Second*30))
+	if metricsManager != nil {
+		options = append(options, withMetrics(metricsManager))
+	}
+	return connect(address, options)
 }
```
gnufied, line 80: I was thinking, a builder pattern might be more suitable for building options to connect. So something like:

```golang
type ConnectionBuilder struct {
	address           string
	metricsManager    metrics.CSIMetricsManager
	timeout           time.Duration
	enableOtelTracing bool
}

func NewConnectionBuilder(address string) *ConnectionBuilder {
	return &ConnectionBuilder{
		address: address,
	}
}

func (b *ConnectionBuilder) WithMetrics(mm metrics.CSIMetricsManager) *ConnectionBuilder {
	b.metricsManager = mm
	return b
}

func (b *ConnectionBuilder) WithTimeout(td time.Duration) *ConnectionBuilder {
	b.timeout = td
	return b
}

func (b *ConnectionBuilder) WithOtelTracing() *ConnectionBuilder {
	b.enableOtelTracing = true
	return b
}

func (b *ConnectionBuilder) Connect() (*grpc.ClientConn, error) {
	return connect(b.address, b.metricsManager,....)
}
```

So a caller would do something like:

```
connectionBuilder := NewConnectionBuilder("unix:////blah").WithMetrics(xxx).WithTimeout(xxx)
connection, err := connectionBuilder.Connect()
```

This might be more easier to extend in future. I do realize, this is new interface functions but it should be fine to use it. We can eventually deprecate both `Connect` and `ConnectWithoutMetrics` in future.

What do you think?

connection/connection.go
```diff
@@ -70,12 +72,11 @@ func SetMaxGRPCLogLength(characterCount int) {
 // For other connections, the default behavior from gRPC is used and
 // loss of connection is not detected reliably.
 func Connect(address string, metricsManager metrics.CSIMetricsManager, options ...Option) (*grpc.ClientConn, error) {
-	return connect(address, metricsManager, []grpc.DialOption{grpc.WithTimeout(time.Second * 30)}, options)
-}
-
-// ConnectWithoutMetrics behaves exactly like Connect except no metrics are recorded.
-func ConnectWithoutMetrics(address string, options ...Option) (*grpc.ClientConn, error) {
-	return connect(address, nil, []grpc.DialOption{grpc.WithTimeout(time.Second * 30)}, options)
```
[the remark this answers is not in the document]
Fricounet: Yes sure, I had not realised a new version was released with this change available so I thought nobody could use it. I will put it back with a deprecation warning.

connection/connection.go
Fricounet: I do not have a strong opinion on this but here are some thoughts:
- I am not a big fan of breaking the interface because it means people will have to update their code even if they do not want to change the behaviour of `Connect`. Considering `csi-lib-utils` is used in all sidecars and some csi-drivers that makes a fair amount of 'clients' that will have to update their code.
- initially, I sticked to the `Options` pattern because I saw a similar patterns in other CSI repos ([metrics in csi-lib-utils](https://github.com/kubernetes-csi/external-resizer/blob/master/vendor/github.com/kubernetes-csi/csi-lib-utils/metrics/metrics.go#L222), [aws-ebs-csi-driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/driver.go#L76), [azuredisk-csi-driver](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/pkg/azuredisk/azuredisk.go#L138), ...) so I felt like sticking to the same pattern was a good idea for code consistency.
- on the other hand, I did not see any other builder pattern in CSI repos (if you have examples, feel free to share them) so it does not feel very consistent with the existing code.

But if you feel like a builder pattern is truly the way to go here, I'm fine with implementing it.
Also, maybe we can ask other folks' opinion on this? Not too familiar with who would be the right person though :/

connection/connection.go
```diff
@@ -105,29 +106,59 @@ func ExitOnConnectionLoss() func() bool {
...
+		o.metricsManager = metricsManager
+	}
+}
+
+// WithOtelTracing enables the recording of traces on the gRPC calls with opentelemetry gRPC interceptor.
+func WithOtelTracing() Option {
+	return func(o *options) {
+		o.enableOtelTracing = true
+	}
+}
```
gnufied, line 128: btw why are `withTimeout` and `withMetrics` unexported but `WIthOtelxxx` is exported?

connection/connection.go
gnufied: okay. I can probably live with function(option) pattern.  I have a minor question below. Another thing  is - we are about to start cutting CSI sidecars. Were you hoping to merge this in sidecars and start adding CLI option?  Do you want to do this after this release or in current release? (I pinged you on slack with same question).

connection/connection.go
Fricounet: Mainly because they were already set when calling `Connect` with the previous implementation and I did not want to let the user override it (because there was no need until now) and if the need were to arise, the functions could simply be exported.
But I can also directly export them and let them override the default I set in `Connect` if you prefer.

connection/connection.go
gnufied: Btw supplying a variadic number of arguments for setting options vs builder pattern are basically two sides of the same coin and not all that much different.  So basically:

```golang
foo = Connect("foo", withBar(), withBaz(), withFoo())

vs

foo = NewConnect("foo").WithBar().WithBaz().WithFoo()

```

I am not a fan of functions that takes too many parameters. Also it may just be me, but those functions that returns functions, only for setting a field in a option struct seems slightly complicated for its own good (it takes a minute for me to parse those), where as builder seems more straightforward.

I agree that we shouldn't break existing interfaces and we can leave `Connect` as it is, but I do tend to prefer the API of builder vs functions that takes options.

connection/connection.go
Fricounet: I exported the function and let them override the default set in `Connect`

connection/connection.go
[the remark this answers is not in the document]
Fricounet: Put the function back :)

Fricounet:
I think I addressed all your comments @gnufied

sunnylovestiramisu:
Can we use a version of go.opentelemetry.io metrics that is compatible among component-base, csi-lib-util and external-provisioner?

Fricounet:
Hey @sunnylovestiramisu, from what I see, the versions used are as follow:
- csi-lib-utils: [v.0.38.0](https://github.com/kubernetes-csi/csi-lib-utils/blob/master/go.mod#L50)
- external-provisioner: [v0.31.0](https://github.com/kubernetes-csi/external-provisioner/blob/master/go.mod#L102)
- component-base: [v0.31.0 ](https://github.com/kubernetes/component-base/blob/master/go.mod#L63)

How do you want to go about it? Personally, I'd prefer all versions to be upgraded to v0.38.0, however I can try to raise a PR to downgrade the version to v0.31.0 in this repo if you prefer?

Fricounet:
Looking at if I can downgrade the version to v0.31.0 here, I fear that won't be possible. The dependency on `go.opentelemetry.io/otel/metric` was added in the `otelgrpc` package in [version v0.37.0](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/instrumentation/google.golang.org/grpc/otelgrpc/v0.37.0/instrumentation/google.golang.org/grpc/otelgrpc/go.mod#L88) but the version used for metric was already `v0.34.0` so I feel like upgrading the dependencies in external-provisioner and component-base is the better solution.

PR #148, opened 2023-09-11
Downgrade otelgrpc lib to resolve dependency conflict when k8s.io/component-base is imported. 

MartinForReal:
**What type of PR is this?**

/kind cleanup

Downgrade otelgrpc lib to resolve dependency conflict when k8s.io/component-base is imported. 

go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.41.0 requires following libs
```
	go.opentelemetry.io/otel v1.16.0
	go.opentelemetry.io/otel/metric v1.16.0
	go.opentelemetry.io/otel/trace v1.16.0
```

where k8s.io/component-base requires a different version

https://github.com/kubernetes/component-base/blob/44f31b29cc46816113a74802aade8f3256483bee/go.mod#L20-L24

The developer needs to resolve dependency conflict(go.opentelemetry.io/otel,go.opentelemetry.io/otel/metric and etc) when k8s/component-base and github.com/kubernetes-csi/csi-lib-utils are imported(e.g. in csi plugin implementation). 

I think it is a good idea to keep go.mod aligned with k8s.io/component-base because k8s.io/component-base is widely used.
If we import  go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.36.2. 
go.opentelemetry.io/otel will not be upgraded to 1.16. And dependency issue can be resolved. 

Fixes #

```release-note
go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc is downgrade to v0.36.2
```


pohly:
> >	go.opentelemetry.io/otel v1.16.0
> >	go.opentelemetry.io/otel/metric v1.16.0
> >	go.opentelemetry.io/otel/trace v1.16.0
>
> where k8s.io/component-base requires a different version

Why is that a conflict? These are v1 packages. Their API should not break between minor or patch releases. If csi-lib-utils pulls in more recent releases of these, then component-base should still work.

It feels like I am missing some piece of explanation.

MartinForReal:
I agree if these dependencies work well when it is invoked in this project( e.g. in ut/e2e)
But when csi driver imports csi-lib-utils, otel and k8s.io/component-base, the developer should choose the correct version of otel which is compatible with csi-lib-utils and k8s.io/component-base. And sometimes it is not possible if otel introduces a break change in new release. So, it is a good idea to keep aligned with config in latest component-base release.  Because there will be less dependency issue in csi driver.

pohly:
Keeping `otelgrpc` the same as in component-base makes sense because it it is a v0 package and thus may have breaking changes. But that's not what your PR description says - it just talks about packages that it pulls in.

MartinForReal:
What I actually mean is Keeping otelgrpc the same as in component-base.

go.mod
```diff
@@ -6,8 +6,8 @@ require (
 	github.com/container-storage-interface/spec v1.8.0
 	github.com/golang/protobuf v1.5.3
 	github.com/stretchr/testify v1.8.2
-	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.41.0
-	go.opentelemetry.io/otel/trace v1.15.0
+	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.36.2
```
pohly, line 9: This is not quite what current Kubernetes master is using:
```
go.mod: go.opentelemetry.io/contrib/instrumentation/github.com/emicklei/go-restful/otelrestful v0.35.0
go.mod: go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.35.0
....
go.mod: go.opentelemetry.io/otel/metric v0.31.0 // indirect
```

If the goal is to be aligned, why not use exactly the same versions?

MartinForReal: updated.

msau42:
/hold

We were able to find a compatible version in https://github.com/kubernetes-csi/external-provisioner/pull/1025

go.mod
MartinForReal: this library is not imported in k8s.io/component-base. but glad to know it is imported in k/k. updated the pr

MartinForReal:
@msau42 yes. We can find one by chance. But It is easier for all of the dependents because there is no need to resolve dependency conflict.

PR #149, opened 2023-09-12
Support structured logging

bells17:
**What type of PR is this?**

/kind feature

I've updated the klog functions used within csi-lib-utils to structured logging functions, following the guidelines below:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md

[Several CSI Sidecars like external-snapshotter rely on csi-lib-utils](https://github.com/kubernetes-csi/external-snapshotter/blob/8c3a30582684a499201d620b7ca3669988d12b9e/go.mod#L11), and to fully support structured logging in these Sidecars, it's necessary for csi-lib-utils to adopt structured logging as well.

In addition, I've modified the log messages based on the following guideline:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#remove-string-formatting-from-log-message

I’ve updated the klog functions in use according to the guidelines provided below, and I've confirmed that they pass the [logcheck](https://github.com/kubernetes-sigs/logtools/tree/main/logcheck) tests:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#change-log-functions

Fixes https://github.com/kubernetes-csi/csi-lib-utils/issues/150

```release-note
Added support for structured logging (the log messages have been changed due to the activation of structured logging)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,3 +18,9 @@ all:
 	go build `go list ./... | grep -v 'vendor'`
 
 include release-tools/build.make
+
+# Check contextual logging.
+.PHONY: logcheck
+test: logcheck
+logcheck:
+	hack/verify-logcheck.sh
--- a/connection/connection.go
+++ b/connection/connection.go
@@ -29,6 +29,8 @@ import (
 	"github.com/kubernetes-csi/csi-lib-utils/protosanitizer"
 	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/backoff"
+	"google.golang.org/grpc/credentials/insecure"
 	"k8s.io/klog/v2"
 )
 
@@ -73,21 +75,21 @@ func SetMaxGRPCLogLength(characterCount int) {
 //
 // For other connections, the default behavior from gRPC is used and
 // loss of connection is not detected reliably.
-func Connect(address string, metricsManager metrics.CSIMetricsManager, options ...Option) (*grpc.ClientConn, error) {
+func Connect(ctx context.Context, address string, metricsManager metrics.CSIMetricsManager, options ...Option) (*grpc.ClientConn, error) {
 	// Prepend default options
 	options = append([]Option{WithTimeout(time.Second * 30)}, options...)
 	if metricsManager != nil {
 		options = append([]Option{WithMetrics(metricsManager)}, options...)
 	}
-	return connect(address, options)
+	return connect(ctx, address, options)
 }
 
 // ConnectWithoutMetrics behaves exactly like Connect except no metrics are recorded.
 // This function is deprecated, prefer using Connect with `nil` as the metricsManager.
-func ConnectWithoutMetrics(address string, options ...Option) (*grpc.ClientConn, error) {
+func ConnectWithoutMetrics(ctx context.Context, address string, options ...Option) (*grpc.ClientConn, error) {
 	// Prepend default options
 	options = append([]Option{WithTimeout(time.Second * 30)}, options...)
-	return connect(address, options)
+	return connect(ctx, address, options)
 }
 
 // Option is the type of all optional parameters for Connect.
@@ -97,7 +99,7 @@ type Option func(o *options)
 // connection got lost. If that callback returns true, the connection
 // is reestablished. Otherwise the connection is left as it is and
 // all future gRPC calls using it will fail with status.Unavailable.
-func OnConnectionLoss(reconnect func() bool) Option {
+func OnConnectionLoss(reconnect func(context.Context) bool) Option {
 	return func(o *options) {
 		o.reconnect = reconnect
 	}
@@ -105,19 +107,25 @@ func OnConnectionLoss(reconnect func() bool) Option {
 
 // ExitOnConnectionLoss returns callback for OnConnectionLoss() that writes
 // an error to /dev/termination-log and exits.
-func ExitOnConnectionLoss() func() bool {
-	return func() bool {
+func ExitOnConnectionLoss() func(context.Context) bool {
+	return func(ctx context.Context) bool {
 		terminationMsg := "Lost connection to CSI driver, exiting"
 		if err := os.WriteFile(terminationLogPath, []byte(terminationMsg), 0644); err != nil {
-			klog.Errorf("%s: %s", terminationLogPath, err)
+			klog.FromContext(ctx).Error(err, "Failed to write a message to the termination logfile", "terminationLogPath", terminationLogPath)
 		}
-		klog.Exit(terminationMsg)
+		klog.FromContext(ctx).Error(nil, terminationMsg)
+		klog.FlushAndExit(klog.ExitFlushTimeout, 1)
 		// Not reached.
 		return false
 	}
 }
 
 // WithTimeout adds a configurable timeout on the gRPC calls.
+// Note that this timeout also prevents all attempts to reconnect
+// because it uses context.WithTimeout internally.
+//
+// For more details, see https://github.com/grpc/grpc-go/issues/133
+// and https://github.com/kubernetes-csi/csi-lib-utils/pull/149#discussion_r1574707477
 func WithTimeout(timeout time.Duration) Option {
 	return func(o *options) {
 		o.timeout = timeout
@@ -139,7 +147,7 @@ func WithOtelTracing() Option {
 }
 
 type options struct {
-	reconnect         func() bool
+	reconnect         func(context.Context) bool
 	timeout           time.Duration
 	metricsManager    metrics.CSIMetricsManager
 	enableOtelTracing bool
@@ -147,22 +155,28 @@ type options struct {
 
 // connect is the internal implementation of Connect. It has more options to enable testing.
 func connect(
+	ctx context.Context,
 	address string,
 	connectOptions []Option) (*grpc.ClientConn, error) {
+	logger := klog.FromContext(ctx)
 	var o options
 	for _, option := range connectOptions {
 		option(&o)
 	}
 
+	bc := backoff.DefaultConfig
+	bc.MaxDelay = time.Second
 	dialOptions := []grpc.DialOption{
-		grpc.WithInsecure(),                    // Don't use TLS, it's usually local Unix domain socket in a container.
-		grpc.WithBackoffMaxDelay(time.Second),  // Retry every second after failure.
+		grpc.WithTransportCredentials(insecure.NewCredentials()), // Don't use TLS, it's usually local Unix domain socket in a container.
+		grpc.WithConnectParams(grpc.ConnectParams{Backoff: bc}),  // Retry every second after failure.
 		grpc.WithBlock(),                       // Block until connection succeeds.
 		grpc.WithIdleTimeout(time.Duration(0)), // Never close connection because of inactivity.
 	}
 
 	if o.timeout > 0 {
-		dialOptions = append(dialOptions, grpc.WithTimeout(o.timeout))
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, o.timeout)
+		defer cancel()
 	}
 
 	interceptors := []grpc.UnaryClientInterceptor{LogGRPC}
@@ -186,19 +200,24 @@ func connect(
 		lostConnection := false
 		reconnect := true
 
-		dialOptions = append(dialOptions, grpc.WithDialer(func(addr string, timeout time.Duration) (net.Conn, error) {
+		dialOptions = append(dialOptions, grpc.WithContextDialer(func(ctx context.Context, addr string) (net.Conn, error) {
+			logger := klog.FromContext(ctx)
 			if haveConnected && !lostConnection {
 				// We have detected a loss of connection for the first time. Decide what to do...
 				// Record this once. TODO (?): log at regular time intervals.
-				klog.Errorf("Lost connection to %s.", address)
+				logger.Error(nil, "Lost connection", "address", address)
 				// Inform caller and let it decide? Default is to reconnect.
 				if o.reconnect != nil {
-					reconnect = o.reconnect()
+					reconnect = o.reconnect(ctx)
 				}
 				lostConnection = true
 			}
 			if !reconnect {
 				return nil, errors.New("connection lost, reconnecting disabled")
+			}
+			var timeout time.Duration
+			if deadline, ok := ctx.Deadline(); ok {
+				timeout = time.Until(deadline)
 			}
 			conn, err := net.DialTimeout("unix", address[len(unixPrefix):], timeout)
 			if err == nil {
@@ -212,14 +231,14 @@ func connect(
 		return nil, errors.New("OnConnectionLoss callback only supported for unix:// addresses")
 	}
 
-	klog.V(5).Infof("Connecting to %s", address)
+	logger.V(5).Info("Connecting", "address", address)
 
 	// Connect in background.
 	var conn *grpc.ClientConn
 	var err error
 	ready := make(chan bool)
 	go func() {
-		conn, err = grpc.Dial(address, dialOptions...)
+		conn, err = grpc.DialContext(ctx, address, dialOptions...)
 		close(ready)
 	}()
 
@@ -231,7 +250,7 @@ func connect(
 	for {
 		select {
 		case <-ticker.C:
-			klog.Warningf("Still connecting to %s", address)
+			logger.Info("Still connecting", "address", address)
 
 		case <-ready:
 			return conn, err
@@ -241,15 +260,14 @@ func connect(
 
 // LogGRPC is gPRC unary interceptor for logging of CSI messages at level 5. It removes any secrets from the message.
 func LogGRPC(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
-	klog.V(5).Infof("GRPC call: %s", method)
-	klog.V(5).Infof("GRPC request: %s", protosanitizer.StripSecrets(req))
+	logger := klog.FromContext(ctx)
+	logger.V(5).Info("GRPC call", "method", method, "request", protosanitizer.StripSecrets(req))
 	err := invoker(ctx, method, req, reply, cc, opts...)
 	cappedStr := protosanitizer.StripSecrets(reply).String()
 	if maxLogChar > 0 && len(cappedStr) > maxLogChar {
 		cappedStr = cappedStr[:maxLogChar] + fmt.Sprintf(" [response body too large, log capped to %d chars]", maxLogChar)
 	}
-	klog.V(5).Infof("GRPC response: %s", cappedStr)
-	klog.V(5).Infof("GRPC error: %v", err)
+	logger.V(5).Info("GRPC response", "response", cappedStr, "err", err)
 	return err
 }
 
@@ -286,14 +304,14 @@ func (cmm ExtendedCSIMetricsManager) RecordMetricsClientInterceptor(
 		if additionalInfo != nil {
 			additionalInfoVal, ok := additionalInfo.(AdditionalInfo)
 			if !ok {
-				klog.Errorf("Failed to record migrated status, cannot convert additional info %v", additionalInfo)
+				klog.FromContext(ctx).Error(nil, "Failed to record migrated status, cannot convert additional info", "additionalInfo", additionalInfo)
 				return err
 			}
 			migrated = additionalInfoVal.Migrated
 		}
 		cmmv, metricsErr := cmm.WithLabelValues(map[string]string{metrics.LabelMigrated: migrated})
 		if metricsErr != nil {
-			klog.Errorf("Failed to record migrated status, error: %v", metricsErr)
+			klog.FromContext(ctx).Error(metricsErr, "Failed to record migrated status")
 		} else {
 			cmmBase = cmmv
 		}
--- a/connection/connection_test.go
+++ b/connection/connection_test.go
@@ -39,6 +39,8 @@ import (
 
 	"github.com/kubernetes-csi/csi-lib-utils/metrics"
 	"k8s.io/component-base/metrics/testutil"
+	"k8s.io/klog/v2/ktesting"
+	_ "k8s.io/klog/v2/ktesting/init"
 )
 
 func tmpDir(t *testing.T) string {
@@ -108,7 +110,8 @@ func TestConnect(t *testing.T) {
 	addr, stopServer := startServer(t, tmp, nil, nil, nil)
 	defer stopServer()
 
-	conn, err := Connect(addr, metrics.NewCSIMetricsManager("fake.csi.driver.io"))
+	_, ctx := ktesting.NewTestContext(t)
+	conn, err := Connect(ctx, addr, metrics.NewCSIMetricsManager("fake.csi.driver.io"))
 	if assert.NoError(t, err, "connect via absolute path") &&
 		assert.NotNil(t, conn, "got a connection") {
 		assert.Equal(t, connectivity.Ready, conn.GetState(), "connection ready")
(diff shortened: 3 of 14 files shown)


bells17:
@pohly Thank you for your review.

> Instead of "just" doing structured logging, would it be possible to also do contextual logging, i.e. remove the dependency on the global logger?
> 
> This makes libraries more versatile.

I've made an attempt to incorporate contextual logging as well. Could you please review it again?
(Sorry for the inconvenience, I had to force push due to a failed conflict resolution.)

I have also confirmed that the following tests with logcheck pass:

```
$ logcheck -check-contextual ./accessmodes/
$ logcheck -check-contextual ./connection/
$ logcheck -check-contextual ./deprecatedflags/
$ logcheck -check-contextual ./leaderelection/
$ logcheck -check-contextual ./metrics/
$ logcheck -check-contextual ./protosanitizer/
$ logcheck -check-contextual ./rpc/
```

connection/connection.go
```diff
@@ -108,9 +108,10 @@ func ExitOnConnectionLoss() func() bool {
 	return func() bool {
 		terminationMsg := "Lost connection to CSI driver, exiting"
 		if err := os.WriteFile(terminationLogPath, []byte(terminationMsg), 0644); err != nil {
-			klog.Errorf("%s: %s", terminationLogPath, err)
+			klog.Background().Error(err, "Failed to write a message to the termination logfile", "terminationLogPath", terminationLogPath)
```
pohly, line 111: Using klog.Background().Error is the same as calling klog.ErrorS: it's structured logging, but not contextual logging.

You need to change the APIs of csi-lib-utils so that it accepts a `logger klog.Logger` or `ctx context.Context` parameter, whatever is more appropriate.

pohly (changes requested):

connection/connection.go
bells17: @pohly I had fix it. Would you please review this PR again?
I've tried configuring it to accepts a `ctx context.Context` parameter because csi-lib-utils is a library used by external packages.

connection/connection.go
```diff
@@ -146,6 +147,7 @@ type options struct {
 
 // connect is the internal implementation of Connect. It has more options to enable testing.
 func connect(
+	ctx context.Context,
```
pohly, line 150: Now that `connect` has a context, `gprc.Dial` should get replaced with `grpc.DialContext`.

connection/connection.go
```diff
@@ -239,15 +241,13 @@ func connect(
 
 // LogGRPC is gPRC unary interceptor for logging of CSI messages at level 5. It removes any secrets from the message.
 func LogGRPC(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
-	klog.V(5).Infof("GRPC call: %s", method)
-	klog.V(5).Infof("GRPC request: %s", protosanitizer.StripSecrets(req))
+	klog.FromContext(ctx).V(5).Info("GRPC call", "method", method, "request", protosanitizer.StripSecrets(req))
```
pohly, line 244: In this function, `FromContext(ctx)` gets called twice and therefore it makes sense to use a `logger := klog.FromContext(ctx)` helper variable.

connection/connection.go
```diff
@@ -188,7 +190,7 @@ func connect(
 			if haveConnected && !lostConnection {
 				// We have detected a loss of connection for the first time. Decide what to do...
 				// Record this once. TODO (?): log at regular time intervals.
-				klog.Errorf("Lost connection to %s.", address)
+				klog.FromContext(ctx).Error(nil, "Lost connection", "address", address)
```
pohly, line 193: In functions with several different log calls, use a `logger := klog.FromContext(ctx)` helper variable.

deprecatedflags/deprecatedflags.go
```diff
@@ -29,19 +31,20 @@ import (
 // and instead using the option triggers a deprecation warning. The
 // return code can be safely ignored and is only provided to support
 // replacing functions like flag.Int in a global variable section.
-func Add(name string) bool {
-	flag.Var(deprecated{name: name}, name, "This option is deprecated.")
+func Add(ctx context.Context, name string) bool {
```
pohly, line 34: I think in this case it is better to pass a `klog.Logger`. It makes the purpose of the parameter clearer and I cannot imagine any scenario where command line flag parsing will need timeouts.

leaderelection/leader_election.go
```diff
@@ -174,14 +174,15 @@ func (l *leaderElection) Run() error {
 		RetryPeriod:   l.retryPeriod,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: func(ctx context.Context) {
-				klog.V(2).Info("became leader, starting")
+				klog.FromContext(l.ctx).V(2).Info("became leader, starting")
```
pohly, line 177: The context is always the same, you can use a `logger` variable.

rpc/common.go
```diff
@@ -133,9 +133,9 @@ func GetGroupControllerCapabilities(ctx context.Context, conn *grpc.ClientConn)
 
 // ProbeForever calls Probe() of a CSI driver and waits until the driver becomes ready.
 // Any error other than timeout is returned.
-func ProbeForever(conn *grpc.ClientConn, singleProbeTimeout time.Duration) error {
+func ProbeForever(ctx context.Context, conn *grpc.ClientConn, singleProbeTimeout time.Duration) error {
 	for {
-		klog.Info("Probing CSI driver for readiness")
+		klog.FromContext(ctx).Info("Probing CSI driver for readiness")
```
pohly, line 138: Helper variable....

pohly (changes requested):

bells17:
TODO: I feel that we might want to consider adding golangci-lint tests to csi-release-tools and incorporating logcheck for klog's structured and contextual logging checks. 
https://github.com/kubernetes-csi/csi-release-tools

connection/connection.go
```diff
@@ -146,21 +149,27 @@ type options struct {
...
+	logger := klog.FromContext(ctx)
 	var o options
 	for _, option := range connectOptions {
 		option(&o)
 	}
 
+	bc := backoff.DefaultConfig
+	bc.MaxDelay = time.Second
 	dialOptions := []grpc.DialOption{
-		grpc.WithInsecure(),                   // Don't use TLS, it's usually local Unix domain socket in a container.
```
bells17, line 157: `grpc.WithInsecure` is deprecated, so it has been changed.

connection/connection.go
```diff
@@ -146,21 +149,27 @@ type options struct {
...
-		grpc.WithBackoffMaxDelay(time.Second), // Retry every second after failure.
```
bells17, line 158: `grpc.WithBackoffMaxDelay` is deprecated, so it has been changed.

bells17:
@pohly Thank you for your review. I've addressed the issues you pointed out.
Could you please review this PR again?

connection/connection.go
```diff
@@ -104,13 +106,14 @@ func OnConnectionLoss(reconnect func() bool) Option {
 
 // ExitOnConnectionLoss returns callback for OnConnectionLoss() that writes
 // an error to /dev/termination-log and exits.
-func ExitOnConnectionLoss() func() bool {
-	return func() bool {
+func ExitOnConnectionLoss() func(context.Context) bool {
```
pohly, line 109: As the comment says, this is for `OnConnectionLoss`. That option has to be changed accordingly, i.e. take a function with a context parameter.

connection/connection.go
```diff
@@ -184,11 +193,11 @@ func connect(
 		lostConnection := false
 		reconnect := true
 
-		dialOptions = append(dialOptions, grpc.WithDialer(func(addr string, timeout time.Duration) (net.Conn, error) {
+		dialOptions = append(dialOptions, grpc.WithContextDialer(func(ctx context.Context, addr string) (net.Conn, error) {
 			if haveConnected && !lostConnection {
 				// We have detected a loss of connection for the first time. Decide what to do...
 				// Record this once. TODO (?): log at regular time intervals.
-				klog.Errorf("Lost connection to %s.", address)
+				logger.Error(nil, "Lost connection", "address", address)
```
pohly, line 200: `logger` here should come from the `ctx` parameter, not the surrounding function.

rpc/common.go
```diff
@@ -133,9 +133,10 @@ func GetGroupControllerCapabilities(ctx context.Context, conn *grpc.ClientConn)
 
 // ProbeForever calls Probe() of a CSI driver and waits until the driver becomes ready.
 // Any error other than timeout is returned.
-func ProbeForever(conn *grpc.ClientConn, singleProbeTimeout time.Duration) error {
+func ProbeForever(ctx context.Context, conn *grpc.ClientConn, singleProbeTimeout time.Duration) error {
```
pohly, line 136: What happens when the context gets canceled? Should `ProbeForever` then return?

That behavior would make sense to me as it would be consistent with other functions accepting a context which then block.

pohly (changes requested):

bells17:
@pohly　Sorry for the delay in responding. I've addressed the issues you pointed out.
Could you please review this PR again?

connection/connection.go
pohly: `OnConnectionLoss` still needs to be updated.

connection/connection.go
pohly: Resolved.

connection/connection.go
pohly: Still open. Add a `logger := klog.FromContext(ctx)`.

rpc/common.go
pohly: Still open.

pohly (changes requested):
Adding logcheck as in the the provisioner lib would be useful.

rpc/common.go
bells17: In case of context cancellation, I have implemented the handling as shown below. Would this implementation be appropriate? I would appreciate your guidance if my understanding is incorrect.

```go
select {
	case <-ctx.Done():
		return nil
...
```

connection/connection.go
bells17: I've also updated the OnConnectionLoss argument. Please let me know if anything is not as intended.

bells17:
@pohly 
> Adding logcheck as in the provisioner lib would be useful.

https://github.com/kubernetes-csi/csi-lib-utils/pull/149#pullrequestreview-1697223351

Based on your suggestion, I have considered the following steps. Would you kindly confirm if they are in line with what you envisaged?

1. Creating a PR to csi-release-tools for adding scripts and Makefile tasks to build the logcheck plugin and run golangci-lint, akin to hack/verify-golangci-lint.sh in kubernetes/kubernetes.
2. Updating csi-release-tools in csi-lib-utils following the PR's merge.
3. Introducing a test in the test-infra repo for csi-lib-utils repo, similar to pull-kubernetes-verify-lint in kubernetes/kubernetes.

Additionally, I would be grateful for your guidance on the following points:

- Would it be appropriate to use a golangci.yaml.in file similar to that in kubernetes/kubernetes for configuring golangci-lint?
- Should we directly use scripts like hack/verify-golangci-lint.sh from kubernetes/kubernetes?

pohly:
> Based on your suggestion, I have considered the following steps. Would you kindly confirm if they are in line with what you envisaged?
>
> 1. Creating a PR to csi-release-tools for adding scripts and Makefile tasks to build the logcheck plugin and run golangci-lint,

Yes. Do it so that "make test" runs the additional check. Use a `.logcheck.conf` in the root for the check. If there is none, fall back to a config in `csi-release-tools` which only checks for correct calls, but doesn't require structured or contextual logging. Hmm, does logcheck even do anything useful then? If not, just don't check anything.

> akin to hack/verify-golangci-lint.sh in kubernetes/kubernetes.

No, that won't be necessary.

> Updating csi-release-tools in csi-lib-utils following the PR's merge.

Yes. You can and should prototype this by editing `release-tools` in `csi-lib-utils` directly and creating a PR. Once it works, move the changes into a commit in `csi-release-tools`.

> Introducing a test in the test-infra repo for csi-lib-utils repo, similar to pull-kubernetes-verify-lint in kubernetes/kubernetes.

No, the normal "test" job for each repo will run the additional check.

> Would it be appropriate to use a golangci.yaml.in file similar to that in kubernetes/kubernetes for configuring golangci-lint?

The specific mechanism from k/k might be overkill for CSI sidecars. A single configuration should be enough.  Just beware that introducing golangci-lint is likely to be more work (but worthwhile!). Better check with SIG Storage first, I don't speak for them.

bells17:
Thank you. I remembered that you had previously advised using logcheck directly in [this discussion](https://github.com/kubernetes-sigs/sig-storage-lib-external-provisioner/pull/154#discussion_r1371581715), so I have attempted to implement this approach. I would be grateful for your thoughts on this.

> Yes. You can and should prototype this by editing release-tools in csi-lib-utils directly and creating a PR. Once it works, move the changes into a commit in csi-release-tools.

Since directly editing release-tools causes the test-subtree task to fail, I have added the logcheck task directly to the Makefile.

Makefile
```diff
@@ -18,3 +18,10 @@ all:
 	go build `go list ./... | grep -v 'vendor'`
 
 include release-tools/build.make
+
+# Check contextual logging.
+.PHONY: logcheck
+test: logcheck
+logcheck:
+	go install sigs.k8s.io/logtools/logcheck@v0.8.1
+	PATH=$$(go env GOPATH)/bin:$$PATH logcheck -check-contextual ./...
```
pohly, line 27: If some developer runs "make test", this will overwrite their `$GOPATH/bin/logcheck`, which will be surprising.

A better solution is to install into a temporary directory (`GOBIN=<tmp dir> go install ...`) and invoke the tool there with an absolute path (no need to set `PATH`).

Remember to delete the temporary directory with `trap EXIT`.

This implies doing everything in a single shell. A verify-logcheck.sh script might be easier than doing it all inline in the Makefile.

bells17: @pohly I have created the `verify-logcheck.sh` script and updated the Makefile to execute this script. For the installation of logcheck, a temp dir is used, and it is set up to be deleted at the end of the script execution using `trap EXIT`.

Could you please review it again? Thank you.@pohly I have created the `verify-logcheck.sh` script and updated the Makefile to execute this script. For the installation of logcheck, a temp dir is used, and it is set up to be deleted at the end of the script execution using `trap EXIT`.

Could you please review it again? Thank you.

connection/connection.go
```diff
@@ -139,30 +142,36 @@ func WithOtelTracing() Option {
...
+		grpc.WithConnectParams(grpc.ConnectParams{Backoff: bc}),  // Retry every second after failure.
 		grpc.WithBlock(),                       // Block until connection succeeds.
 		grpc.WithIdleTimeout(time.Duration(0)), // Never close connection because of inactivity.
 	}
 
 	if o.timeout > 0 {
-		dialOptions = append(dialOptions, grpc.WithTimeout(o.timeout))
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, o.timeout)
+		defer cancel()
```
pohly, line 174: Note to myself: this works because `connect` waits for dialing to finish or fail. Otherwise returning would cause an on-going dial attempt to be aborted.

pohly: There is one problem that was already in the original code: as described in https://github.com/grpc/grpc-go/issues/133, `WithTimeout` also prevents all future attempts to reconnect. That might be surprising when someone wants to use `Connect` with a `OnConnectionLoss` callback that doesn't abort.

Can you add a comment to `WithTimeout` explaining that setting a timeout prevent reconnecting?

connection/connection.go
```diff
@@ -186,20 +195,25 @@ func connect(
...
 				}
 				lostConnection = true
 			}
 			if !reconnect {
 				return nil, errors.New("connection lost, reconnecting disabled")
 			}
+			timeout := 0 * time.Second
+			if deadline, ok := ctx.Deadline(); ok {
+				timeout = time.Until(deadline)
+			}
```
pohly, line 216: I suspect that `net.DialTimeout(..., 0)` will not work. It might never occur, but perhaps switch between `net.Dial` and `net.DialTimeout` depending on whether there is a deadline?

deprecatedflags/deprecatedflags.go
```diff
@@ -29,28 +30,29 @@ import (
 // and instead using the option triggers a deprecation warning. The
 // return code can be safely ignored and is only provided to support
 // replacing functions like flag.Int in a global variable section.
-func Add(name string) bool {
-	flag.Var(deprecated{name: name}, name, "This option is deprecated.")
+func Add(logger klog.Logger, name string) bool {
```
pohly, line 33: What logger would a caller pass in here? Logging hasn't been initialized yet, so it could only pass `klog.Background`, which is already what the code below uses via `klog.Warningf`.

Logging during command line passing never works well because of that. I suggest to replace `klog.Warningf` with `fmt.Fprintf(os.Stderr, ...`

hack/verify-logcheck.sh
```diff
@@ -0,0 +1,82 @@
...
+ensure-temp-dir() {
+  if [[ -z ${CSI_LIB_UTIL_TEMP-} ]]; then
+    CSI_LIB_UTIL_TEMP=$(mktemp -d 2>/dev/null || mktemp -d -t csi-lib-utils.XXXXXX)
+    trap_add cleanup-temp-dir EXIT
+  fi
+}
+
+# Example:  trap_add 'echo "in trap DEBUG"' DEBUG
+# See: http://stackoverflow.com/questions/3338030/multiple-bash-traps-for-the-same-signal
+trap_add() {
```
pohly, line 51: This seems overkill for this script. How about simply this instead of `ensure-temp-dir` below:
```
CSI_LIB_UTIL_TEMP=$(mktemp -d 2>/dev/null || mktemp -d -t csi-lib-utils.XXXXXX)
trap 'rm -rf ${CSI_LIB_UTIL_TEMP}' EXIT
```

rpc/common.go
```diff
@@ -133,30 +133,36 @@ func GetGroupControllerCapabilities(ctx context.Context, conn *grpc.ClientConn)
...
-				return fmt.Errorf("CSI driver probe failed: %s", err)
-			}
-			// Timeout -> driver is not ready. Fall through to sleep() below.
-			klog.Warning("CSI driver probe timed out")
-		} else {
-			if ready {
-				return nil
+		select {
+		case <-ctx.Done():
+			return nil
```
pohly, line 143: The original code did not return on timeout, it kept trying. I think you should return an error here to indicate that probing timed out.

```suggestion
			return ctx.Err()
```

rpc/common.go
```diff
@@ -133,30 +133,36 @@ func GetGroupControllerCapabilities(ctx context.Context, conn *grpc.ClientConn)
...
+		case <-ticker.C:
+			logger.Info("Probing CSI driver for readiness")
+			ready, err := probeOnce(conn, singleProbeTimeout)
```
pohly, line 146: probeOnce needs to use the context and the timeout.

connection/connection.go
bells17: added: https://github.com/kubernetes-csi/csi-lib-utils/pull/149/commits/de384e6e6810061fcb4541102822f8ca8a3cc03d#diff-b9adde9612439aa9df1ceebc36325654ae9d11a48c508b63995688809c1a1115R173-R175

connection/connection.go
bells17: I think that using net.DialTimeout(..., 0) as mentioned above is not problematic, so I've not implemented a switch to net.Dial when timeout=0. I'd appreciate your thoughts on this.
(On a different note, but it seemed to me that using var timeout time.Duration is more readable than timeout := 0 * time.Second, so I have made that change as well.)

deprecatedflags/deprecatedflags.go
bells17: changed: https://github.com/kubernetes-csi/csi-lib-utils/pull/149/commits/d0c5c66abd0ac95a8f5d62e2b36a3beaa3cc5f59#diff-8a66eb22e577246eefa65671fa6c956c8e869e15f6961a8faf8d32ebbd168110L55-R53

hack/verify-logcheck.sh
bells17: changed: https://github.com/kubernetes-csi/csi-lib-utils/pull/149/commits/d0c5c66abd0ac95a8f5d62e2b36a3beaa3cc5f59#diff-2f76c847ffab450d5e5be791a39758261423800a823839cff7dcace87e135b26R29-R32

rpc/common.go
bells17: https://github.com/kubernetes-csi/csi-lib-utils/pull/149/commits/d0c5c66abd0ac95a8f5d62e2b36a3beaa3cc5f59#diff-5fb7c6e20b635d2fb705b5f7ec97d8f1a35c802d9bd76105b814b43ee4885c6eL143-R143

rpc/common.go
bells17: I've made the changes as below, but could you please confirm if the modifications are correct as per your understanding?

bells17:
@pohly Thank you for the review.
I've made revisions following the comments you provided. I would appreciate it if you could confirm the changes.

connection/connection_test.go
```diff
@@ -380,6 +391,10 @@ func TestExplicitReconnect(t *testing.T) {
 }
 
 func TestConnectMetrics(t *testing.T) {
+	contextFromKtesting := func() context.Context {
+		_, ctx := ktesting.NewTestContext(t)
+		return ctx
+	}
```
pohly, line 397: This is unnecessary: all test cases get a new Context, but it's all for the same testing.T. You get the same effect by creating just one Context here and then using it for each test case.

If this was using sub-tests, then sharing the same context would not enable proper per-sub-test output. But it doesn't. so a single context is fine.

rpc/common.go
```diff
@@ -133,36 +133,42 @@ func GetGroupControllerCapabilities(ctx context.Context, conn *grpc.ClientConn)
...
-				return fmt.Errorf("CSI driver probe failed: %s", err)
-			}
-			// Timeout -> driver is not ready. Fall through to sleep() below.
-			klog.Warning("CSI driver probe timed out")
-		} else {
-			if ready {
-				return nil
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
pohly, line 143: Better use this:
```suggestion
			return context.Cause(ctx)
```

If the context gets canceled using the new(is) `WithCancelCause`, then we get that cause, otherwise the normal "context canceled".

rpc/common.go
pohly: See https://github.com/kubernetes-csi/csi-lib-utils/pull/149/files#r1584285702

pohly (commented):
Some small nits, but overall this looks good.

bells17:
@pohly Thank you for providing the feedback and suggestions for improvement.
I have made the revisions, so kindly review the changes.

connection/connection.go
bells17: @pohly Sorry for my misunderstanding. I have removed the comment.

Regarding this:
> I meant adding the comment to `func WithTimeout(...)`. Then it becomes part of the API documentation.

I'm not fully understanding the implication, so I would like to clarify the meaning.
Is this suggesting that a comment should be added to the following page?
https://pkg.go.dev/context#WithTimeout

If my understanding above is correct, please let me know if there's anything else I should do to clarify or address this.

pohly: No. I meant adding it to https://github.com/kubernetes-csi/csi-lib-utils/blob/f82f9de5b8aeb3c3b236d7f58fc5eeab34438078/connection/connection.go#L120-L125

pohly:
/lgtm

@bells17: can you squash into one commit?

bells17:
@pohly Thank you for the review and lgtm!
I've squashed the commits into one.

pohly:
/lgtm
/assign @jsafrane 

For approval.

PR #167, opened 2024-05-07
Test logcheck

bells17:
Debugging our CI
/hold


jsafrane:
/close 
We updated logcheck to 0.8.2 that works well with go 1.22

PR #175, opened 2024-05-30
Restore ProbeForever behavior of probing first before sleeping

Fricounet:
**What type of PR is this?**

/kind bug

The previous beheavior of ProbeForever was to try a probe and then sleep for the `probeInterval` duration. When it was modified to use a `time.Ticker` instead of a sleep in https://github.com/kubernetes-csi/csi-lib-utils/pull/149, it changed the behavior to wait the `probeInterval` first and then do the probe. This PR restores the previous behavior by running the probe first because it is causing issues when trying to start some of the sidecars (https://github.com/kubernetes-csi/external-attacher/pull/561). And in general, it slows down the startup of the sidecars by at least 1s, even when the plugin is already ready to accept connections.
Based on https://stackoverflow.com/a/54752803.

Fixes #

```release-note
Restored the previous behavior of `ProbeForever` to do a probe first before sleeping.
```


jsafrane:
I've tagged v0.18.1, I (or dependabot) will update all sidecars.

PR #184, opened 2024-11-23
cleanups and refactor of protosanitizer

huww98:
**What type of PR is this?**
/kind cleanup

>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR does a series cleanups and refactors about protosanitizer:
- Update the referenced CSI spec package to 1.10
- Remove all the reference to old github.com/golang/protobuf package
- Refactor the protosanitizer to use the new protoreflect package
- Sync the Makefile for csitest.proto with csi spec repo

About the output change of protosanitizer:
- All kinds of fields now supported: including those in array, in map, or in oneof.
- Enum now outputs their name instead of number, I think this is more suitable for logging purpose.
- The nesting level of `oneof` is reduced by one.
Example diff of the output (manually formatted for diff):
```diff
--- old.json
+++ new.json
@@ -12,22 +12,18 @@
     "secrets": "***stripped***",
     "volume_capabilities": [
         {
-            "AccessType": {
-                "Mount": {
-                    "fs_type": "ext4",
-                    "mount_flags": [
-                        "flag1",
-                        "flag2",
-                        "flag3"
-                    ]
-                }
-            },
             "access_mode": {
-                "mode": 5
+                "mode": "MULTI_NODE_MULTI_WRITER"
+            },
+            "mount": {
+                "fs_type": "ext4",
+                "mount_flags": [
+                    "flag1",
+                    "flag2",
+                    "flag3"
+                ]
             }
         }
     ],
-    "volume_content_source": {
-        "Type": null
-    }
+    "volume_content_source": {}
 }
```

Also added a benchmark to justify the new implementation:
Old:
```
goos: darwin
goarch: arm64
pkg: github.com/kubernetes-csi/csi-lib-utils/protosanitizer
cpu: Apple M2 Pro
BenchmarkStrip             98264             11801 ns/op            7488 B/op        149 allocs/op
BenchmarkStripLarge           22          49160320 ns/op        39795573 B/op     620137 allocs/op
```

New:
```
BenchmarkStrip            141199              7608 ns/op            4816 B/op         89 allocs/op
BenchmarkStripLarge          122          12642834 ns/op         9955752 B/op     140009 allocs/op
```

The new implementation is significantly faster, and allocates less memory

Fixes #

protosanitizer now imports csi spec directly, instead of copy over the `e_CsiSecret`. Since this module already requires the csi spec module, this should not break anyone.

Support for CSI spec v0.3 is dropped, since the `csi.pb.go` file in that version does not implement the new protoreflect interface. And again, since this module already requires the csi spec module v1, users cannot use this module with v0.3 even without this PR. So, this is not a new breakage.

```release-note
The output format of protosanitizer is slightly refined.
```


huww98:
@jsafrane Can you also take a look at this one? It should speedup several times for cases like https://github.com/kubernetes-csi/external-provisioner/pull/1310 when the log level is high.

protosanitizer/protosanitizer.go
```diff
@@ -48,67 +48,68 @@ type stripSecrets struct {
...
+	case protoreflect.MessageKind:
+		return stripMessage(v.Message())
+	case protoreflect.EnumKind:
+		return field.Enum().Values().ByNumber(v.Enum()).Name()
+	default:
+		return v.Interface()
 	}
+}
+
+func stripValue(field protoreflect.FieldDescriptor, v protoreflect.Value) any {
```
jsafrane, line 77: Can be `stripValue()` and `stripSingleValue()` combined into a single function with a bigger switch?

```go
switch {
case field.IsList():
...

case field.IsMap():
...

case field.Kind() == protoreflect.MessageKind:
...
case field.Kind() == ...

default: return v.Interface()
```

protosanitizer/protosanitizer.go
```diff
@@ -48,67 +48,68 @@ type stripSecrets struct {
...
+	if field.IsList() {
+		l := v.List()
+		res := make([]any, l.Len())
+		for i := range l.Len() {
+			res[i] = stripSingleValue(field, l.Get(i))
```
jsafrane, line 82: For example, what happens when a list has a list? Will it call `stripSingleValue()` with a list, which then leads to `default:` branch there and thus print the list with potential secrets?

protosanitizer/protosanitizer.go
huww98: It is possible, but less clear, I think. Because for field that `IsList()`, its `Kind()` still returns the kind of list element. So, we will have multiple case statements matched.

My code structure actually looks very like the official encoder:
https://github.com/protocolbuffers/protobuf-go/blob/b98563540c0a4edb38526bcd6e6c97f9fac1f453/encoding/prototext/encode.go#L201-L212

protosanitizer/protosanitizer.go
huww98: List is represented as repeated field in the protobuf. Like:
```proto3
  repeated VolumeCapability volume_capabilities = 3;
```
And it cannot be nested without going through a level of message. i.e.:
```proto3
  repeated repeated VolumeCapability volume_capabilities = 3;
```
is not valid, but this one is:
```proto3
  repeated Topology accessible_topology = 5;

message Topology {
  map<string, string> segments = 1;
}
```

Similarly, `map<string, map<string, string>>` or `repeated map<string, string>` are not valid.

huww98:
https://github.com/golang/protobuf/issues/1655 We may get official redact support in the future. But even after that is ready, I think we should still keep ours, because that one will be prototext based, not json.

protosanitizer/protosanitizer.go
jsafrane: Thanks for the explanation!
/lgtm
/approve

PR #185, opened 2024-12-17
Update kubernetes to 1.32

dfajmon:
**What type of PR is this?**
/kind cleanup

Update to the latest release-tools
Bump Kubernetes to v1.32.0

```release-note
Update Kubernetes dependencies to 1.32.0
```


huww98:
Do we really need to update the deps so often? This is a lib, and does not produce executable itself. Bumping dep version forces every consumer to update transitive deps when updating this module, which is unnecessary. I think we should only update the deps when we really depend on the new feature. Consumers can still use newer version than declared in this repo.

dfajmon:
@huww98 We update libraries once for release unless there are bug fixes that might need more bumps. It's also better to have a unified Kubernetes version across all projects in one release.

huww98:
> It's also better to have a unified Kubernetes version across all projects in one release.

I agree with this for modules that produce executable, and rarely imported by other projects. But for module like this repo, updating has no benefits, I think. Golang will pick the updated version specified in the main module anyway.

Our project is currently using client-go 1.30.0, and I just want to use the new protosanitizer (`1 file changed, 56 insertions(+), 113 deletions(-)`). And that `go mod vendor` brings me `1981 files changed, 60843 insertions(+), 117187 deletions(-)` to my vendor dir, which is far more than necessary. Yes, we will upgrade eventually. But I think it is not necessary to force every project in the community to follow tightly. At least allow us to use the still maintained backport version?

PR #193, opened 2025-04-25
Add package standardflags for -automaxprocs commandline argument

nixpanic:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

The new -automaxprocs flag can be used to set the GOMAXPROCS environment variable to match the configured Linux container CPU quota.

Related: kubernetes-csi/node-driver-registrar#495

Applications can call EnableAutomaxprocs() to enable it during startup.

```release-note
The AddAutomaxprocs() function adds the -automaxprocs boolean flag to the commandline options. By default the flag is disabled.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/stretchr/testify v1.10.0
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.58.0
 	go.opentelemetry.io/otel/trace v1.33.0
+	go.uber.org/automaxprocs v1.6.0
 	google.golang.org/grpc v1.69.0
 	google.golang.org/protobuf v1.36.0
 	k8s.io/api v0.32.0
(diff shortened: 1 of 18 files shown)


xing-yang:
What are the requirements for using this flag properly?

nixpanic:
> What are the requirements for using this flag properly?

The flag only has effect when the container where the application runs has [CPU limits](https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/#specify-a-cpu-request-and-a-cpu-limit) configured.

The [maxprocs](https://pkg.go.dev/go.uber.org/automaxprocs/maxprocs) package reads the CPU quota from cgroups under `/sys/...` and sets the `GOMAXPROCS` environment variable based on the CPU quota. If CPU quota is not configured, or not supported, the flag does not have any effect.

